PR #3, opened 2021-01-17
EOS R stubs update for firmware 1.8.0

kitor:


kitor:
I'd squash those commits.

PR #6, opened 2021-02-21
initial m50.110 support

b-:
wow, very merge


reticulatedpines:
This got merged indirectly!  Kitor took pieces from yours and updated from other work, and tested on another M50.  Thanks for the help!

PR #12, opened 2021-06-03
add CONFIG_EOSRP definitions

coon42:
prepare full ML port by adding CONFIG_EOSRP defines (copied from R) and adding dynamic range table.


coon42:
obsolete since integrated into #13 .

PR #19, opened 2021-08-17
EOSRP: adjust RESTARTSTART value

coon42:
Value was too low and lead to a crash on boot. This PR increases the value which fixes the issue.


coon42:
invalid. follow up in #21

PR #24, opened 2021-10-16
RP: display shutter and electronic shutter count

coon42:
This PR enables the features `FEATURE_SHOW_SHUTTER_COUNT` and `FEATURE_SHOW_TOTAL_SHOTS` in `EOSRP.160/features.h`. These settings add a menu entry to the Debug menu which displays the mechanical and the electronic shutter count.

The RP seems to have a bug in its firmware as the shutter count is extremely high when directly read. However, the value seems have the same offset on all RPs. At the forums we've got a readout from a fresh RP with one photo taken: https://www.magiclantern.fm/forum/index.php?topic=22770.msg232297#msg232297 Therefore we expect the offset being always `1086947309`, so it is subtracted from the shutter count value.


reticulatedpines:
Merged via cherry-pick

PR #36, opened 2021-11-27
Updates for 750D, M50, R180

kitor:


kitor:
Already FF merged into dev, thus closing PR.

Issue #30, opened 2021-11-13
LV overlays issues for D678 Photo mode


kitor:
General issues (Digic 6)
* `D6..` `winsys_sem` (`Semaphore_Winsys_v` aka vsync semaphore) DNE on 750, but it exists on 5D4 (and `D.78X`)
  * Usage in `D6..` removed in bae1c85d185e1850ad795caa7c5f10fdf106b4b8
  * Without semaphore, screen tearing is possible while redrawing ML GUI.
  * TODO: Investigate when it was introduced by Canon, adapt code accordingly.

General issues (Digic 6 and up):
* `D678` Even when no Canon overlays are visible, Canon will draw some on actions (eg. shutter speed change)
  * Canon overlays will flicker as we actively overdraw whole screen while `zebra_should_run()` (94c0e0325235e437fb707a57bf5a2acc1d1d661e)
  * To be investigated: possibility to kill Canon GUI drawing in this mode
    * Guess: this was achieved by `bmp_on()` / `bmp_off()` on D45
    * In `FEATURE_COMPOSITOR_XCM` we could just mute Canon layer(s), but there should be more generic solution.
* `D678`: HTP can have multiple levels now. Adapt to same style display as ALO.

General issues (Digic 8):
* `D..8` `PROP_LENS_DYNAMIC_DATA` / `lens` struct size issues
   * RAW values (and field sizes) for ISO, AV, TV had changed. 
* `D..8` `CONFIG_COMPOSITOR_DEDICATED_LAYER` implementation for EOS RP and up (code to setup our own layer)
  * R code is specific to R; RP, 250D and even R6 (Digic X) seems to have unified code
  * Different structures are used to setup GUI layers
  * Can be probably achieved using just XCM function calls, without touching `RENDERER` / `VMIX` structs.
  * Refer to Chapter 4 for clues: [Compositors, layers, contexts in RGB and YUV - How Digic 7(6?)+ draw GUI](https://www.magiclantern.fm/forum/index.php?topic=26024) 
* `D..8` Most dial modes are displayed properly except Fv + SCN + Movie Mode are displayed as ?

Partially fixed:
* `D..8` 338dd8efe13de19ec8844bf90cea0cde75f13ddd Detection if overlays are clear to draw is broken: 
  * TODO: check possibility to improve delay by handling props from 958c07fdd8cb36399d62eda309f7145d589db620
* `D678` Many (not yet enabled) code paths search for dialog struct in a broken way
[...]

PR #38, opened 2021-12-15
A ton of D678 updates.

kitor:
Addresses many issues from #30 and more.

Needed changes:
Models added in meantime require `LiveViewApp_dialog` stub (data pointer). Easy to find, just check where return pointer from `CreateDialogBox()` for `LiveViewApp_handler` is saved. On other models I already prepared (but commented out) `*_dialog` pointers for other apps we will need in future - but those are not needed right now.

For Digic 8 models with XCM and more than one Canon layer it is worth enabling `CONFIG_COMPOSITOR_XCM` (requires `XCM_GetSourceSurface()` and `_pXCM` stubs). This eliminates race condition on boot where regular code can get pointer to wrong (focus overlays) layer, making our rendering broken (invisible).

### DIFF
--- a/src/compositor.c
+++ b/src/compositor.c
@@ -1,5 +1,5 @@
 /** \file
- * VRAM RGBA Compositor interface
+ * VRAM Compositor interface
  */
 /*
  * kitor: So I found a compositor on EOSR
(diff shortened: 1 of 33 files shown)


src/zebra.c
```diff
@@ -3556,21 +3556,33 @@ static void draw_zoom_overlay(int dirty)
...
+        CURRENT_GUI_MODE <= 3 &&
+        job_state_ready_to_take_pic() &&
+        !mirror_down;
+#else
     struct gui_task * current = gui_task_list.current;
     struct dialog * dialog = current->priv;
     extern thunk LiveViewApp_handler;
 
     #if defined(CONFIG_5D3) || defined(CONFIG_M50)
     extern thunk LiveViewLevelApp_handler;
```
kitor, line 3603: Changes below in that function just restore code to a state prior to initial D6+ modifications.

src/compositor.c
```diff
@@ -29,52 +29,120 @@
...
+ * refreshVrmsSurface (renamed to VMIX_TransferRectangleToVram on Digic X)
+ * is used by Canon code to render Input layers into XOC.
+ *
+ * On pure XIMR cameras it uses XIMR functions to setup Ximr Context
+ */
+extern void      refreshVrmsSurface();
+extern uint32_t  display_refresh_needed;
 
+// Store our Input Layer ID.
 int _rgb_vram_layer = 0;
```
reticulatedpines: I guess the layer IDs should be consts, LAYER_CANON or something

src/compositor.c
```diff
@@ -29,52 +29,120 @@
...
+
+    uint16_t bmp_w = BMP_W_PLUS - BMP_W_MINUS;
+    uint16_t bmp_h = BMP_H_PLUS - BMP_H_MINUS;
+
+    // fill MARV data
+    pNewLayer->signature    = 0x5652414D;  // MARV
+    pNewLayer->bitmap_data  = pBitmapData;
+    pNewLayer->opacity_data = 0x0;
+    // 0x5040100 is used by D78. Some early D6 use a different value
+    pNewLayer->flags        = 0x5040100;   // bitmask (?) for RGBA
```
reticulatedpines: Better to make this a constant if you already know cams differ, easier for people to find and fix later

src/compositor.c
kitor: Not sure how one would name it. Can't be `CANON_LAYER` as depending on model there may be one or more Canon layers. `LAYER_NOT_ALLOCATED` is kind of true, but on the other hand if code fails to allocate our own, it deliberately stays at layer "0", falling back to behavior like XCM was not enabled.

src/compositor.c
kitor: Yes, probably should be a constant with a caveat - this is model dependent. At the moment I don't know any EOS codebase model that would use "old" IDs.

I'll create a constant in a follow up commit.

reticulatedpines (approved):

src/compositor.c
reticulatedpines: Discussed in discord, fine for now

src/compositor.c
kitor: Addressed in 26e9d97dc1573a79b51500d5c1c60719d4e7f6d5

src/compositor.c
kitor: I addressed it in 26e9d97dc1573a79b51500d5c1c60719d4e7f6d5

coon42 (approved):

src/zebra.c
```diff
@@ -3556,21 +3556,33 @@ static void draw_zoom_overlay(int dirty)
 
 int liveview_display_idle()
 {
+#ifdef CONFIG_DIGIC_678
+    extern struct dialog* LiveViewApp_dialog;
+
+/**
+ * kitor TODO: contains a lot of duplicates from D45 code.
+ * Do we want to keep it that way? Seems cleaner to me.
+ */
```
reticulatedpines: Since there's a big shared block and the test is always "&&" with the other checks, you could have this early on and not need it later in the function:

if       !(LV_NON_PAUSED &&
        DISPLAY_IS_ON &&
        !menu_active_and_not_hidden() &&
        gui_state == GUISTATE_IDLE &&
        CURRENT_GUI_MODE <= 3 &&
        job_state_ready_to_take_pic() &&
        !mirror_down)
    return 0

Now you know any code later on already passed those checks.

src/zebra.c
```diff
@@ -3898,6 +3907,32 @@ BMP_LOCK (
     }
 #endif
 
+#ifdef CONFIG_DIGIC_678
+/**
+ * kitor: On D678 apps handlers that are in our interest either doesn't show up
+ * on `gui_task_list` at all, or are buried down on the list (would require
+ * list search via gui_task->next to find those that do).
```
reticulatedpines, line 3924: Any idea about why?  Haven't investigated this myself so don't know how hard it would be, but my pref is to keep close to the old way - can we always search and continue to use the handlers, rather than this new dialog way?

src/dialog.h
```diff
@@ -159,207 +159,118 @@ struct dialog_list
...
-    uint32_t field_0xfc;
-    uint32_t field_0x100;
-    uint32_t field_0x104;
-    uint32_t field_0x108;
-    uint32_t field_0x10c;
+#if defined(CONFIG_DIGIC_VI) || defined(CONFIG_DIGIC_VII)
+    uint32_t                 unk_d67_01;                // Not initialized in CreateDialogBox; type may be wrong
+#endif
+    struct subscriberClass * hTerminateSubscriber;      // disp_sw_controller?
+    uint32_t                 refresh_x;                 // Region to redraw? probably a "damaged" region.
```
reticulatedpines, line 173: Personally, I really don't like this style of massive whitespace between type and name.  Aligning blocks of addresses can have some value when you want to compare them.  Aligning names doesn't add anything.

Also, the bizarre "type * name" for pointers; I've never seen a codebase that uses this before; why add ambiguity with multiplication?

Since this is partially personal taste and both styles are used a lot in this code I'm not suggesting you bother changing this, just giving my opinion on C.  I think ML would benefit from a coding standard that was applied automatically, I don't care too much what that standard is.

src/lens.c
```diff
@@ -2695,19 +2695,45 @@ static LVINFO_UPDATE_FUNC(picq_update)
...
+ *  RAW + l 04020700   X.. .... ..X. .... .XXX .... ....
+ * CRAW + L 03030700   .XX .... ..XX .... .XXX .... ....
+ * CRAW + l 03020700   .XX .... ..X. .... .XXX .... ....
+ */
+        int raw = pic_quality & 0x600; // 3 for *RAW, 0 - not RAW
+        int jpg = (pic_quality & 0x100);
+        int rawsize = pic_quality >> 26; // if raw: 0 CRAW, 1 RAW
+        int jpegtype = (pic_quality >> 16) & 0xF;
+        int jpegsize = pic_quality & 0xFF;
+#else
```
reticulatedpines, line 2795: Many of these are shifted by 8 bits for D8 vs other (only rawsize that isn't?).  You could make the ifdef do the shift, and keep the rest of the code the same.  As is, slightly less compiled code.  If you conditionally shift, might be clearer to read?

src/lens.c
```diff
@@ -1791,6 +1793,81 @@ void _prop_lv_lens_request_update()
      */
     prop_request_change(PROP_LV_LENS, &lv_lens_raw, 0);
 }
+#endif
+
+#ifdef CONFIG_DIGIC_VIII
+#define DYNAMIC_FLAG_ST2_MF    0x80  // true -> MF, false -> AF
+#define DYNAMIC_FLAG_ST3_IS    0x0F  // looks PROP_LV_LENS_STABILIZE equiv.
```
reticulatedpines: These two seem to be used as masks, not flags.  "Dynamic" is presumably because these relate to LENS_DYNAMIC_DATA, not because this mask is dynamic?  Feels confusing as the first word in the name.  st2 and st3 are bad names but I guess these come from readid shell command?  "IS" is confusing in an uppercase context.

Maybe MANUAL_FOCUS_MASK and IMAGE_STAB_MASK?  Could replace MASK with BIT.

src/property.h
```diff
@@ -310,8 +310,10 @@
 #define ALO_HIGH 2
 #define ALO_OFF 3
 
-#if defined(CONFIG_5D3)
+// was guarded for 5D3 / 6D, but R series seem to use it too.
```
reticulatedpines, line 321: I'm not sure on this change.  *Every* define in the 5D3 block is in the 6D one.  The intent may have been to have a block per cam (but...  there are some other 5D3 blocks earlier, and a lot of other mess).  You've now exposed it to all cams.  Might be fine?

This whole file feels like a mess and should probably be split up.

src/zebra.c
kitor: Yep, makes sense. I'll rewrite that.

src/zebra.c
kitor: IMO it was always z-index ordered linked list.

In early D4 code each app had basically just one dialog - or at least this was the case for situations we need to detect. Later it started to fall aparat - just look at this code from `zebra.c` / `liveview_display_idle()`:
```
#if defined(CONFIG_5D3) || defined(CONFIG_M50)
extern thunk LiveViewLevelApp_handler;
#elif defined(CONFIG_DIGIC_V)
extern thunk LiveViewShutterApp_handler;
#endif

#if defined(CONFIG_6D)
extern thunk LiveViewWifiApp_handler;
#endif
```

As you can see from code above, and its earlier D678 adaptations that were substituted with this new implementation - it started to get ridiculous, requiring sometimes complicated per model ifdefs.
For example on EOSR (and only EOSR) topmost app is usually... one that processes touchbar inputs.

IMO right now Canon starts multiple GUI apps running in parallel (eg. that touchbar app is on top on EOS R). Not confirmed theory is that "LvApp" is now just a low level app, with no windows visible in normal run - while everything visible is managed by other dedicated apps. Which would make sense considering increasing complexity.

src/dialog.h
kitor: Yep, I just mimicked how other structures were already formatted around that file. I'd also want to have automatic code formatting, as from experience this just cuts all unnecessary discussions.

src/lens.c
kitor: Yes, `DYNAMIC_` prefix as it applies to `LENS_DYNAMC_DATA`. `ST2` and `ST3` in fact are field names in structure. 

`lens_info.IS               = (_dynamic->st3 & DYNAMIC_FLAG_ST3_IS);`

Uppercase IS is already used in ML owned `lens_info` structure. So while confusing, this + it being a define were the factors why I used that as a part of this name.

Proposed names make much more sense, however I'd still prefix them with something so it is clear those apply to `LENS_DYNAMIC_DATA`

src/property.h
kitor: This exposes only property ID. To use it, you need to enable a code path separately by using `GENERIC_GET_HTP` in `cfn.c` in camera platform dir. Due to that I don't see any downsides of doing this. 

Worst case we can do `ifdef 5D3 / 6D / D678`.

src/lens.c
kitor: In fact everything except `rawsize` seems to be shifted by 8. `rawsize` is probably moved, but with just two options and MSBs not used for anything it can't be assumed that it is still 8 bits.

Conditionally shifting would still require ifdef for that shift + ifdefs for `rawsize`, I'm not sure if this is worth it?

src/zebra.c
reticulatedpines: Okay, I understand it better at least.  Sounds reasonable for why old way isn't great.  Probably cleaner would be rewrite all cams with new method (or searching maybe)?  If you agree, please raise an issue for that future work.

src/dialog.h
reticulatedpines: Yeah.  I thought it was probably that, diffs for review don't show much surrounding code.

src/lens.c
reticulatedpines: I'm not sure either, use whichever you think is cleaner.

src/lens.c
reticulatedpines: lens_info.IS is not too bad because there's no confusion with the rest of the name, it's the whole name.

Oh and can't use BIT, my brain was confused by 0x80 and saw 0xf as a single bit.

If you can think of a nice name where the prefix doesn't make it crazy long, go for it.

src/zebra.c
kitor: Yes, it needs some research for older models, but so far I'm nearly sure that this one can be backported. It also removes null pointer reads that are possible with current `gui_task->next` implementation.

src/property.h
reticulatedpines: Yes, I think I'm just being cautious because I don't understand why they were in blocks to start with, when there are no conflicting definitions.  Defines don't use any space.  I don't understand the purpose of the per cam blocks, so it triggers my Chesterton's Fence alarm.

src/lens.c
kitor: Then I'll stay as is. Old way has no documentation / proof dumped from real camera - we just have to trust the code it is ok. New has all the examples put in comment, verified R/RP/M50. If old one would be documented well, I'd reconsider - but just for that I'd keep them separated.

src/zebra.c
kitor: Addressed in 9b38e7696a5b9961307a6fbac10caa8ad9258899

src/lens.c
kitor: I decided to drop defines at all. They are overkill in this case, made sense during development when I had multiple those defined for some fields.

See a87a19e0f53237e8111e32b39a693a6c683b7756 for update.

src/zebra.c
kitor: I added more description in `9b38e7696a5b9961307a6fbac10caa8ad9258899` as `liveview_display_idle()` had similar modification.

As for issue for future, I already have this in #30 in "Partially fixed" list (_Many (not yet enabled) code paths search for dialog struct in a broken way_). I'll probably create a separate issue for that later, as this new method still needs to be proven of other cases than LV detection (PlayApp, etc).

src/lens.c
reticulatedpines: Only used in one place and fairly obvious patterns for masks, I'm okay with this.

PR #39, opened 2021-12-26
RP.160: Enable ML LiveView overlay

coon42:
This PR adds the required changes to enable ML liveview overlays for EOS RP.160.
As `autoexec.bin` got bigger, `ML_RESERVED_MEM` has been increased and `RESTARTSTART` value is adjusted so it fits into RAM again.

Additionally:

- `EVF_STATE` state object address was added and invalid addresses were removed from `internals.h`
- `SRM` is disabled for now as it is not stable / not implemented properly yet.
- `XCM` is enabled
- finally `FEATURE_GLOBAL_DRAW` + `CONFIG_LIVEVIEW` + `CONFIG_STATE_OBJECT_HOOKS` is enabled to display ML overlay in liveview

Working Things:
- Clock is displayed properly
- Image Quality setting is displayed properly for all possible permutations of `RAW/CRAW/L/l/M/m/S1/s1/S2`
- Focal length is displayed properly when zooming in and out (tested over the full range with a RF24-240 lens)
- IS (Image Stabilization) is displayed properly. In B (Bulb) mode, it is displayed in cyan color.
- `AF` (Auto Focus) and `MF` (Manual Focus) are displayed properly
- `Alo` (Automatic Light Optimizer) is displayed properly for all possible settings: low = `alo`, Standard = `Alo`, High = `ALO`
- `HTP` (Highest Tone Priority) is displayed when set to `D+` or `D+2`.
- Battery level is correct for all three battery levels RP provides: Full (White with all bars), Half Full (Yellow) Empty (Red).

(Probably) Not workig Things:
~~- `Alo` (Automatic Light Optimizer?) is always displayed (not sure how this works yet)~~
- Most dial modes are displayed properly except `Fv` + `SCN` + `Movie Mode` are displayed as `?`
- Aparture value is wrong (Always displays `f5.6` or no value at all)
- ISO value is wrong (Always displayed as `AUTO`)


kitor:
> Most dial modes are displayed properly except Fv + SCN + Movie Mode are displayed as ?

Added to #30 

> Aparture value is wrong (Always displays f5.6 or no value at all)
> ISO value is wrong (Always displayed as AUTO)

Expected, already in #30. It is also true for Tv (shutter speed).

> Alo (Automatic Light Optimizer?) is always displayed (not sure how this works yet)

`cfn.c` in platform dir should handle that (copy content from e.g. R). All D678 so far use generic ALO/HTP handlers.

kitor (approved):

coon42:
> `cfn.c` in platform dir should handle that (copy content from e.g. R). All D678 so far use generic ALO/HTP handlers.

Copied `cfn.c` from R.180 and now it is displayed properly. However:

> HTP (Highest Tone Priority) is displayed when set to D+ or D+2.

Maybe we should follow the convention of `Alo` here and display `Htp` for `D+` and `HTP` for `D+2` instead here?

PR #53, opened 2022-01-14
Fix ptpcam build on macos

fabianfreyer:


contrib/ptpcam/ptpcam.c
```diff
@@ -3349,7 +3349,7 @@ int chdk(int busn, int devn, short force)
                         }
                         
                         rate = 1000000.0f / (avgDelta / maxVal);
-                        printf("Overflows: %i �sec (avg: %f), maxVal = 0x%08X, tickRate = %f Hz\n", delta, avgDelta, maxVal, rate);
+                        printf("Overflows: %i sec (avg: %f), maxVal = 0x%08X, tickRate = %f Hz\n", delta, avgDelta, maxVal, rate);
```
reticulatedpines, line 3352: Why remove this?  It's the correct character for micro-seconds

contrib/ptpcam/ptp.c
```diff
@@ -26,6 +26,7 @@
 #include <stdarg.h>
 #include <stdio.h>
 #include <string.h>
+#include <unistd.h>
```
reticulatedpines, line 29: I'd guess this breaks builds on Windows - has this been tested?  (I'm not 100% sure it ever worked on Windows, if old code never worked there this is okay)

coon42: According to https://github.com/reticulatedpines/magiclantern_simplified/blob/dev/contrib/ptpcam/Makefile.win32 I don't think this builds on windows out of the box as it uses a hard coded path to a precompiled libusb lib file, which the user has to provide. Instead of depending to the lib we should probably provide the libusb sources instead (maybe via sub module?) and make it use WSL2 instead of cygwin.

We may also update to a newer libusb version, as the currently used one is pretty old.

As the PR doesn't introduce any issues on Linux, I'd suggest to merge this.

reticulatedpines: Ah, thanks.  That makefile looks like our source is intended to build on Windows, although as you say probably with some binary people have to find.  It looks like cygwin can (does?) provide unistd.h, so this may build with the change.  I don't mind too much if we break it, if anyone is using it they can complain.

If we want to support Windows itself, we shouldn't use WSL; that's just supporting a different flavour of Linux.

contrib/ptpcam/ptpcam.c
fabianfreyer: Ah, I think on my system it broke because unicode is not valid C... I guess this should be then "usec" or correctly escaped.

reticulatedpines: Got a ref for unicode not being valid C?  My belief is it's allowed and implementation defined.  What error were you seeing, with what toolchain?  The ptpcam makefile forces gcc, I would take a guess this is a gcc implementation detail and you're using a Mac that has gcc aliased to clang?  Unicode escaping is from C99 onwards, not sure what we'll default to.

Changing to "usec" is fine by me.

reticulatedpines:
Closing: merged the early commits and modified the last commit to use "usec".

PR #55, opened 2022-01-19
R6 initial stubs

lclevy:
should we merge anyway for better debug collaboration ?
the readme states clearly this code is freezing


reticulatedpines:
The stubs look useful, but the commit history is quite messy.  Cbasic stuff isn't required, LED blink tests and some changes to other files that will change behaviour of non-R6 cams.

Can this be rewritten to be a smaller number of clean commits that makes only the necessary changes to get good stubs, consts, and early boot behaviour for R6?

reticulatedpines:
Closed by 890073910bf1ec5a43ab0546c0ff045290751f9e

PR #58, opened 2022-02-18
bmp: don't provide invalid handle to XCM_GetSourceSurface

coon42:
This fixes spamming `[WINSYS] ERROR XCM_GetSourceSurface Illegal handle` when booting a camera in factory mode without having display attached where XCM gets not initialized.

### DIFF
--- a/src/bmp.h
+++ b/src/bmp.h
@@ -67,7 +67,7 @@ inline uint8_t *rgb_vram_preinit()
 {
 #ifdef CONFIG_COMPOSITOR_XCM
     // Get address of MARV for layer 0 from XCM
-    rgb_vram_info = XCM_GetSourceSurface(_pXCM, 0);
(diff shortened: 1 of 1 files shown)


kitor:
Use case is invalid as ML requires GUI to boot (it will wait for vram pointer in a loop). But this change may reduce a number of winsys messages on regular boot.

reticulatedpines:
Do we want this change?  The comments make me unsure.  How would I test this on cams with display attached?  It should work unchanged compared to before I guess?

kitor:
We discussed that on Discord and the conclusion was that ML is not suited to run in "factory mode" anyway so it makes no sense, as whole Magic Lantern code is dependent on having working display/drawing routines.

In a normal run this change will silence a few prints from that function (it will print error until XCM is initialized).

coon42:
We probably shouldn't merge this. I Will just use this locally when booting into factory mode.

PR #59, opened 2022-02-19
RP.160: enable CHDK PTP commands

coon42:
This PR enables CHDK PTP commands for RP.160. This allows interacting with the camera via `ptpcam` tool. The following CHDK commands were testes successfully yet:

- download
- upload
- m (read memory)

All other commands are untested and not recommended to be used, once they have verified to work.

### DIFF
--- a/platform/RP.160/Makefile.platform.default
+++ b/platform/RP.160/Makefile.platform.default
@@ -23,4 +23,4 @@ PLATFORM_ARCH = armv7-a
 #endif
 
 ML_BOOT_OBJ     = boot-d678.o
-ML_SRC_EXTRA_OBJS += function_overrides.o test_features.o
+ML_SRC_EXTRA_OBJS += function_overrides.o test_features.o ptp.o ptp-chdk.o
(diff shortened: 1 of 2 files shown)


reticulatedpines:
Does it make sense to make this the default config on upstream?  Feels like it should be some build option maybe, since nothing else uses it.

coon42:
I mainly want to use this to update `autoexec.bin` via PTP without having to swap my SD card all the time. This works totally fine but not all `ptpcam` functions are fully tested yet, so I can't tell that it is 100% safe yet.

I agree that having some kind of build option could be a solution here, as PTP gets acticated as soon as `ptp.c` and `ptp-chdk.c` are compiled and linked in.

reticulatedpines:
Can something be done with this?  It's three years old.

reticulatedpines:
Closed due to inactivity.  Feel free to re-open if you still want to work on this.

It is probably better to work in a branch on your fork, then submit the PR when it is ready for merging.

PR #60, opened 2022-05-24
90d initial stubs

MakersFunDuck:
Compared stubs in 850d  with 90d, and updated possible ones from 90d firmware. Needs validation!


reticulatedpines:
Closing because we have someone else working on 6d2 now, and that port is further along.

PR #63, opened 2022-06-14
Add Github Actions workflow

petabyt:
A Github action workflow to compile ML for all models, and compile all modules.

### DIFF
--- a/.github/workflows/makefile.yml
+++ b/.github/workflows/makefile.yml
@@ -2,9 +2,9 @@ name: Build Magic Lantern
 
 on:
   push:
-    branches: [ gh-actions ]
+    branches: [ dev ]
   pull_request:
-    branches: [ gh-actions ]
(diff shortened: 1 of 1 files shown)


reticulatedpines:
It's interesting, but I don't currently have any intention of using GH actions.  I'd rather not depend on GH specific functionality.  If there's some reason why closing this will cause you difficulties, let me know and we can work something out, otherwise I'll close this in probably a week or so.

petabyt:
It's OK if you close it. It was mainly a personal experiment to see if it would work. Also, `.github/workflows` is Github specific and shouldn't cause any incompatibilities if the files were moved over to hg or other code forges. The actions run completely separately from the Git repository.

petabyt:
I had some ideas to extend this workflow, (maybe support Gitlab, convert to shell script). Let me know if you have any questions or regain interest in this.

PR #69, opened 2022-09-01
Add basic tetris to dev

petabyt:
A few notes:
- "ptetris" is a minimal tetris library I made. It interfaces with a common "binding" to IO. https://github.com/petabyt/ptetris/
- Code doesn't recompile when ptetris.h is modified
- Needs better field checking. Currently doesn't check out of field after rotating
- `menu_redraw_blocked` is set to `1` to prevent tetris rendering to flicker as menu is being drawn.
- `task_create` is called to keep the ML menus in place after the game is quit.


modules/tetris/Makefile
```diff
@@ -0,0 +1,7 @@
+# Force format
+$(info $(shell astyle --style=allman tetris.c ptetris.h))
```
reticulatedpines: This looks like it adds a dependency to the ML build system for "astyle", since all modules are built for all cams.  I don't really want to do that.  If you're volunteering to convert all ML code to be compatible, and *then* add an astyle dep, that's different!  But it also feels like a separate task.

Probably remove this dep?

modules/tetris/ptetris.h
```diff
@@ -0,0 +1,302 @@
...
+#include <stdio.h>
+#include <string.h>
+
+void pt_pixel(int x, int y, int col);
+
+extern uint32_t pt_colors[];
+
+// Max array size
+#ifndef PT_MAX_WIDTH
+	#define PT_MAX_WIDTH 40
```
reticulatedpines, line 10: This file uses tabs, the .c uses 4-spaces.  4-spaces is preferred, please convert this file.

modules/tetris/Makefile
petabyt: I can comment it out.

modules/tetris/ptetris.h
petabyt: This is a copied directly from my ptetris project, which uses tabs. So I would have to convert the file to spaces every time I update the file. Not a big deal.

modules/tetris/Makefile
reticulatedpines: Thanks.  Annoyingly, there's an "official" coding style in doc/CODING_STYLE - which even suggests using astyle!  But it is very clear this is not followed consistently.

petabyt: Would be nice if the entire repo was formatted consistently, but that would be too big of a diff and mess with history+branches

reticulatedpines:
Test merge worked.  Bugs found:
 - pieces can spawn outside of the boundary, and then they cannot be moved into play.  Seems to happen if piece was dropped at either edge and cursor left there, some pieces then spawn incorrectly
 - square block never spawns, so presumably something around the RNG is incorrect
 - sleeploop variable is never referenced and should be removed (triggers a compiler warning)
 - the left- and right-handed S pieces are the same colour!  It hurts my brain!

It's a cool mini-tetris :)

petabyt:
Some of those issues seem familiar. I've got an updated ptetris.h that I never pushed, I'll add it later

petabyt:
Fixed all of the things you mentioned. IMO it's very playable at this point. Could still use some minor improvement (better visuals, better controls) but for tetris on a DSLR it's not bad :)

PR #75, opened 2022-10-20
Add "configure script" (WIP)

petabyt:
- Python script that checks system dependencies
- Helps avoid random bugs, avoids mysterious errors
- rst2html, pip2, python2, arm gcc correct version, bash, host gcc
- Spews a lot of garbage output in console currently

Any suggestions welcome, I hardly understand the ML build system.

### DIFF
--- a/configure.py
+++ b/configure.py
@@ -1,11 +1,11 @@
 import os, platform, subprocess
 
-def checkCommand(shell):
+def check_command(shell):
     return subprocess.check_call(["which", shell])
 
-def checkAll():    
+def check_all():    
     print("Checking for rst2html")
-    if checkCommand("rst2html"):
+    if check_command("rst2html"):
         print("rst2html not found")
         print("Install with: pip3 install rst2html")
         exit(1)
@@ -13,22 +13,22 @@ def checkAll():
         print("rst2html checked")
 
     print("Checking for python2")
-    if checkCommand("python2"):
+    if check_command("python2"):
         print("python2 not found. ML still relies on it.")
         exit(1)
     else:
         print("python2 works")
 
     print("Checking for host GCC")
-    if checkCommand("gcc"):
+    if check_command("gcc"):
         print("gcc not found")
         exit(1)
     else:
         print("gcc checked")
 
     print("Checking for ARM GCC")
-    if checkCommand("arm-none-eabi-gcc"):
+    if check_command("arm-none-eabi-gcc"):
         print("Could not find arm-none-eabi-gcc")
         exit(1)
 
-checkAll()
+check_all()
--- a/configure.py
+++ b/configure.py
@@ -1,20 +1,9 @@
 import os, platform, subprocess
 
 def checkCommand(shell):
-    if platform.system() == "Windows":
-        return os.system("where " + shell)
-    else:
-        return os.system("which " + shell)
+    return subprocess.check_call(["which", shell])
 
-def checkAll():
-    print("Checking for valid bash")
-    if checkCommand("bash"):
-        print("bash not found")
-        print("You will want a POSIX compatible environment")
-        exit(1)
-    else:
-        print("bash checked")
-    
+def checkAll():    
     print("Checking for rst2html")
     if checkCommand("rst2html"):
         print("rst2html not found")
@@ -30,18 +19,6 @@ def checkAll():
     else:
         print("python2 works")
 
-    print("Checking for pip2")
-    if checkCommand("pip2"):
-        print("pip2 not found. ML still relies on it.")
-        exit(1)
-    else:
-        print("python2 works")
-        py1 = subprocess.check_output(["python2", "--version"])
-        py2 = subprocess.check_output(["python", "--version"])
-        if py1 == py2:
-            print("python is python2")
-        else:
-            print("You will need to route python to python2, you can install 'python-is-python2'")
     print("Checking for host GCC")
     if checkCommand("gcc"):
         print("gcc not found")
@@ -53,11 +30,5 @@ def checkAll():
     if checkCommand("arm-none-eabi-gcc"):
         print("Could not find arm-none-eabi-gcc")
         exit(1)
-    else:
-        version = subprocess.check_output(["arm-none-eabi-gcc", "--version"])
-        if "20160919" in str(version):
-            print("Compiler is correct version for ML")
-        else:
-            print("This version of ARM GCC isn't fully supported, you may run into issues compiling")
 
 checkAll()
--- a/configure.py
+++ b/configure.py
@@ -1,0 +1,63 @@
+import os, platform, subprocess
+
+def checkCommand(shell):
+    if platform.system() == "Windows":
+        return os.system("where " + shell)
+    else:
+        return os.system("which " + shell)
+
+def checkAll():
+    print("Checking for valid bash")
+    if checkCommand("bash"):
+        print("bash not found")
+        print("You will want a POSIX compatible environment")
+        exit(1)
+    else:
+        print("bash checked")
+    
+    print("Checking for rst2html")
+    if checkCommand("rst2html"):
+        print("rst2html not found")
+        print("Install with: pip3 install rst2html")
+        exit(1)
+    else:
+        print("rst2html checked")
+
+    print("Checking for python2")
+    if checkCommand("python2"):
+        print("python2 not found. ML still relies on it.")
+        exit(1)
+    else:
+        print("python2 works")
+
+    print("Checking for pip2")
+    if checkCommand("pip2"):
+        print("pip2 not found. ML still relies on it.")
+        exit(1)
+    else:
+        print("python2 works")
+        py1 = subprocess.check_output(["python2", "--version"])
+        py2 = subprocess.check_output(["python", "--version"])
+        if py1 == py2:
+            print("python is python2")
+        else:
+            print("You will need to route python to python2, you can install 'python-is-python2'")
+    print("Checking for host GCC")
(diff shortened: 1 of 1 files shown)


petabyt:
Configure scripts in Python seem to be common: https://github.com/ninja-build/ninja/blob/master/configure.py

configure.py
```diff
@@ -0,0 +1,63 @@
...
+            print("You will need to route python to python2, you can install 'python-is-python2'")
+    print("Checking for host GCC")
+    if checkCommand("gcc"):
+        print("gcc not found")
+        exit(1)
+    else:
+        print("gcc checked")
+
+    print("Checking for ARM GCC")
+    if checkCommand("arm-none-eabi-gcc"):
```
reticulatedpines: In theory, multiarch GCC can build ML, and that is plain "gcc".  Untested by me

configure.py
```diff
@@ -0,0 +1,63 @@
+import os, platform, subprocess
+
+def checkCommand(shell):
+    if platform.system() == "Windows":
+        return os.system("where " + shell)
```
reticulatedpines: Prefer subprocess, with params passed as a list, e.g.:
subprocess.run(["ls", "-l"])

This removes problems with shell and is the standard way to run stuff from python.  You can also use variants such as subprocess.check_call() if that helps with error handling.

configure.py
```diff
@@ -0,0 +1,63 @@
...
+        print("pip2 not found. ML still relies on it.")
+        exit(1)
+    else:
+        print("python2 works")
+        py1 = subprocess.check_output(["python2", "--version"])
+        py2 = subprocess.check_output(["python", "--version"])
+        if py1 == py2:
+            print("python is python2")
+        else:
+            print("You will need to route python to python2, you can install 'python-is-python2'")
```
reticulatedpines: I thought I fixed this by removing all plain "python" usage, making all calls in the build system python2 or python3, and/or making all #!/usr/bin/env lines explicit python2 or python3.

Notably, my system doesn't have "python", and it works fine, so it will fail this configure check.

configure.py
```diff
@@ -0,0 +1,63 @@
...
+    else:
+        print("gcc checked")
+
+    print("Checking for ARM GCC")
+    if checkCommand("arm-none-eabi-gcc"):
+        print("Could not find arm-none-eabi-gcc")
+        exit(1)
+    else:
+        version = subprocess.check_output(["arm-none-eabi-gcc", "--version"])
+        if "20160919" in str(version):
```
reticulatedpines: My build env reports this, which would fail this check:
arm-none-eabi-gcc (15:8-2019-q3-1+b1) 8.3.1 20190703 (release) [gcc-8-branch revision 273027]

configure.py
```diff
@@ -0,0 +1,63 @@
...
+
+    print("Checking for python2")
+    if checkCommand("python2"):
+        print("python2 not found. ML still relies on it.")
+        exit(1)
+    else:
+        print("python2 works")
+
+    print("Checking for pip2")
+    if checkCommand("pip2"):
```
reticulatedpines: My system has no pip2 and builds fine

configure.py
```diff
@@ -0,0 +1,63 @@
...
+
+def checkCommand(shell):
+    if platform.system() == "Windows":
+        return os.system("where " + shell)
+    else:
+        return os.system("which " + shell)
+
+def checkAll():
+    print("Checking for valid bash")
+    if checkCommand("bash"):
```
reticulatedpines: Modern macOS systems don't install bash.  They might still have bash in path, aliased to something else, not sure, untested by me.

configure.py
petabyt: Couldn't figure out how to get return codes from subprocess when I wrote that. os.system was easier.

configure.py
petabyt: I thought the 2016 version was the only one supported?

configure.py
petabyt: I think I saw a few bash scripts somewhere in the build system

configure.py
petabyt: Don't the python2 scripts need some packages?

configure.py
reticulatedpines: I'm building with the version listed above.

configure.py
reticulatedpines: Quite probably, but I don't have "pip2" so as is this check will fail my build environment.  If we need package X and Y available to python2, the appropriate check will be using python2 to import those packages.  This could be done as a one-liner per package, checking for success.

That would have the bonus of telling people what packages they need to install.

petabyt: So a python2 script that is called by python3?

configure.py
reticulatedpines: I'm not aware of any - did it error when you didn't have bash installed?

configure.py
reticulatedpines: No, a direct check for the existence of the pre-requisite.  E.g., if we depend on the "os" module, this, which will succeed:
python3 -c 'import os'

Compare with a hypothetical "ossss" module, which will fail:
python3 -c 'import ossss'

It's not pip2 that we care about.  It's python2 (or 3), having access to certain modules.  So check for that directly.

configure.py
petabyt: Not sure how I can temporarily remove bash without bricking my os.

reticulatedpines: In that case I'd say remove the check - we don't have any evidence it's required.  If we find someone hits this problem we can document the circumstances and add it back (or fix the unexpected bash dependency).

configure.py
petabyt: I was thinking of a try/catch usage to check if the imports failed.

reticulatedpines: Sure, so was I, something like this:
```
try:
    subprocess.check_call(["python3", "-c", "import ossss"])
except CalledProcessError:
    # can't import ossss, handle this somehow
```

petabyt: I was thinking:
```
try:
    import ossss
except CalledProcessError:
    # can't import ossss, handle this somehow
```

configure.py
petabyt: I can check for `sh` instead, sound good?

configure.py
reticulatedpines: That feels a bit worse?  If you're running as python2 you can't test python3 imports, and vice versa.  My way any version of python can test imports for any other version.

configure.py
reticulatedpines: I don't know.  I think if we don't know what shell we may or may not require, how can we know if we're running a good test?

Probably better to not test this at all, until we see an error from someone and can identify incorrect shell as being the cause.

configure.py
petabyt: I guess so, for python2 a separate file would have to be made.

petabyt:
Updated, still doesn't test modules. Don't think it's a big deal though, it's mostly standard library packages right?

configure.py
reticulatedpines: Resolving: changed to subprocess.check_call()

configure.py
reticulatedpines: Resolving: unnecessary req check for python being python2 was removed

configure.py
reticulatedpines: Resolving: version check was removed.  But see earlier comment: it is in theory possible to build ML without any arm-none-eabi-gcc binary.  Plain gcc, if it's multiarch, can in theory do this.

configure.py
reticulatedpines: Do you want to mark this one resolved?  The pip2 check isn't done anymore, but we're not checking for the modules.  Up to you.

configure.py
reticulatedpines: Resolving: check for bash was removed

configure.py
```diff
@@ -1,20 +1,9 @@
...
-        print("bash not found")
-        print("You will want a POSIX compatible environment")
-        exit(1)
-    else:
-        print("bash checked")
-    
+def checkAll():    
     print("Checking for rst2html")
     if checkCommand("rst2html"):
         print("rst2html not found")
```
reticulatedpines, line 9: The advice in the following line to use pip3 is inappropriate.  rst2html is called from modules/readme2modulestrings.py, which is python2

configure.py
```diff
@@ -0,0 +1,63 @@
+import os, platform, subprocess
```
reticulatedpines, line 1: configure.py is probably not the best name.  People may assume it configures the build.

Perhaps check_build_requirements.py?

configure.py
```diff
@@ -0,0 +1,63 @@
...
+
+def checkCommand(shell):
```
reticulatedpines: This should be check_command(), to meet PEP8 naming conventions.  Same for checkAll()

configure.py
petabyt: Sure

configure.py
petabyt: That would be confusing, maybe check.py

petabyt: I'm thinking in the future, the script might do some configuring

reticulatedpines: Totally fine to rename it to configure if it plays that role in the future, that makes sense.

"check" is okay, a bit generic (what is it checking?).  Currently there is nothing in the file that says what it is for, so all people have is the name.  A good description in the file (argparse and --help?  Or a nice comment) would help.

configure.py
petabyt: fixed

reticulatedpines:
This has stalled I think, and the build system has changed significantly since then, anyway - no longer needs python2.

PR #87, opened 2023-02-20
Pull request - add comment to modules build

Xhyub:
Pull request to add comment in the modules relevant Makefile. It is an issue when copy_modules_with_satisfied_deps.py expects a complete list in "ML_MODULES" and user has not set their module names accordingly.

### DIFF
--- a/modules/Makefile
+++ b/modules/Makefile
@@ -35,6 +35,7 @@ clean: $(MODULES_CLEAN_ALL_TARGETS)
 
 install: | prepare_install_dir
 install: all
+	# set names for modules in Makefile.modules.default
 	$(PYTHON3) copy_modules_with_satisfied_deps.py $(INSTALL_MODULES_DIR) $(ML_MODULES)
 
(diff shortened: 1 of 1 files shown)


reticulatedpines:
Hi - I don't understand the comment in the commit.  Are you saying that you tried to make a new module, and didn't add it to ML_MODULES via modules/Makefile.platform.default?

In that case, yes, the module will not be built.  But will a comment in a Makefile help people understand the problem?  I wouldn't expect it to, they won't see it.

Xhyub:
That was my suggestion, there is no documented rules for building modules in the repository. Where should we direct users to if they wanted to build their own modules and don't know how?

reticulatedpines:
We could definitely do with better documentation.  This is always tricky.  You need someone with good understanding of the whole project to write good docs.

For this PR, I think an experienced C dev wouldn't need the comment, and an inexperienced one wouldn't find it (and would probably have a lot of other problems anyway as they learn the source).

PR #93, opened 2023-04-13
Python3 support

ilia3101:
This eliminates python2 as a dependency for building ML. Module info screens appear to be generating correctly, just as before (which is the primary purpose of the python build scripts). Tested on macos and 5D3.

Main changes:
- Python 3 syntax changes
- Set `PYTHON=python3` in Makefile.user.default
- Updated html2text to a later version, for python3 compatibility (https://github.com/aaronsw/html2text)

I also changed `readme2modulestrings` to call python modules directly, instead of using terminal commands and pipes.
- Calls html2text module directly now
- Calls the docutils module directly, instead of checking possible aliases for rst2html (a utility provided by the docutils module anyway)
- Use re module for the regular expression part instead of grep
- No more pipes.

Still needed:
- Test on windows (make sure it works, confirm module info screens look correct)
- Test on linux
- Check if non-building-related python2 scripts need attention
- Clean things up, combine the `PYTHON` and `PYTHON3` variables in `Makefile.user.default`, etc...

### DIFF
--- a/Makefile.user.default
+++ b/Makefile.user.default
@@ -176,7 +176,7 @@ CCACHE=ccache
 GREP=grep
 MV=mv
 AWK=awk
-PYTHON=python2
+PYTHON=python3
 PYTHON3=python3
 RST2LATEX=rst2latex.py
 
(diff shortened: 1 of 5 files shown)


ilia3101:
Closing, did a better job in https://github.com/reticulatedpines/magiclantern_simplified/pull/95

PR #95, opened 2023-04-13
Remove Python 2 dependency

ilia3101:
Minimal changes this time.

Turns out it wasn't necessary to update html2text, it just needed a few syntax tweaks for Python 3.

Integer division was also taken care of.

### DIFF
--- a/Makefile.user.default
+++ b/Makefile.user.default
@@ -176,7 +176,7 @@ CCACHE=ccache
 GREP=grep
 MV=mv
 AWK=awk
-PYTHON=python2
+PYTHON=python3
 PYTHON3=python3
 RST2LATEX=rst2latex.py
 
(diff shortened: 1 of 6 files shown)


branylagaffe:
Should also the python2 dependencies to create the features' table also be handled, OR, the script isn't used any more ?

reticulatedpines:
> Should also the python2 dependencies to create the features' table also be handled, OR, the script isn't used any more ?

That isn't used in builds for cams, so the urgency is lower.  Currently, we don't generate the feature matrix from this repo - if we bring that back we can fix it then.

reticulatedpines:
Merged here: https://github.com/reticulatedpines/magiclantern_simplified/commit/8e3c5eeca4e8eff0211687c6618431b4a83ffb4c

I also deleted some unused files requiring python2.

Thanks for the help!

PR #96, opened 2023-04-22
RP.160: Add stub for ErrForCamera_handler + DRYOS_ASSERT_HANDLER + increase ML_RESERVED_MEM

coon42:
Those changes are required to make `CONFIG_CRASH_LOG` define work.

### DIFF
--- a/platform/RP.160/consts.h
+++ b/platform/RP.160/consts.h
@@ -37,6 +37,8 @@
 #define MALLOC_STRUCT 0x2A030
 #define MALLOC_FREE_MEMORY (MEM(MALLOC_STRUCT + 8) - MEM(MALLOC_STRUCT + 0x1C)) // "Total Size" - "Allocated Size"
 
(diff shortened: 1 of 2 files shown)


reticulatedpines:
Seems uncontroversial to me.  Did you intend to enable crash log in this PR and forgot?  Either way is fine.

coon42:
Yes, I left it out intentionally in this PR as it doesn't work anyway yet due to a canon bug: Assert Handler calls in RPs 1.6.0 firmware are mostly set to NULL which leads to a crash in `Debug_Assert` function, as it expects a string with a description here. Thus the assert function crashes before our handler is called so a dump is never written to SD card.

However, this can be fixed by implementing a modified Debug Assert function on our own which checks the first param for being a nullptr and providing a default string in that case. By remapping it via MMU, canon firmware will use our fixed assert function which shouldn't crash anymore. However, this will be done in a future PR.

Once Assert handler is fixed, `CONFIG_CRASH_LOG` can be enabled.

PR #119, opened 2023-11-26
Fixes to intervalometer and SRM memory

jmheder:
DISPLAY_IS_ON fixed
SRM_BUFFER_SIZE fixed
Constants  and features  for intervalomater added
Intervalometer was tested with PROP_REQUEST_CHANGE + is_prop_alloweD(...) {return true;}

### DIFF
--- a/platform/7D2.112/consts.h
+++ b/platform/7D2.112/consts.h
@@ -25,6 +25,7 @@
 /* "Malloc Information" */
 #define MALLOC_STRUCT 0x41b50                    // from get_malloc_info, helper of malloc_info
 #define MALLOC_FREE_MEMORY (MEM(MALLOC_STRUCT + 8) - MEM(MALLOC_STRUCT + 0x1C)) // "Total Size" - "Allocated Size"
+#define SRM_BUFFER_SIZE  0x2314000   /* print it from srm_malloc_cbr */
 
 /* high confidence */
 #define DRYOS_ASSERT_HANDLER 0x28d50               // from debug_assert function, hard to miss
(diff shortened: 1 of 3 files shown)


platform/7D2.112/internals.h
```diff
@@ -32,3 +28,7 @@
 #define CONFIG_NEW_TASK_STRUCTS
 #define CONFIG_TASK_STRUCT_V2
 #define CONFIG_TASK_ATTR_STRUCT_V3
+
+// This camera has BULB mode
+#define CONFIG_BULB
+#define CONFIG_SEPAEATE_BULB_MODE // has "B" mode
```
reticulatedpines: Should be SEPARATE not SEPAEATE?

jmheder: Yes, typo. Making the correct define,  solved an issue I had. Using dail "B" mode without setting bulb timer duration  would not make the camera fire, with the corrected define, it it now fired in "B" mode without setting the time

jmheder:
Should be fine now.

platform/7D2.112/consts.h
```diff
@@ -58,7 +59,7 @@
 
 #define CANON_SHUTTER_RATING 200000
 
-    #define DISPLAY_IS_ON               0x1  // TODO: find real value
+#define DISPLAY_IS_ON ((MEM(0x027e98) & 0x100) ? 0 : 1)  // found at 0xfe0b6824 : MEM(0x027e98) == 0x10001(on) 0x10101(off)
```
reticulatedpines: This should probably be:
(!(char)(MEM(0x27e99))) ?

See fe0b65ee(int display_state), which is called with either 0 or 1.

jmheder: Ok, done and tested

reticulatedpines:
This has been integrated as part of my effort to get raw video fully working on 7D2.  Thanks for the help!

PR #201, opened 2025-06-23
Added GitHub Action and Makefile/all

robbcocco:
Added a GitHub Action capable of building and releasing the project.
Also fixed the makefile in platform/all

### DIFF
--- a/.github/workflows/build_and_publish.yml
+++ b/.github/workflows/build_and_publish.yml
@@ -1,0 +1,135 @@
+name: Build & Publish Magic Lantern
+
+on:
+  workflow_dispatch:
+    inputs:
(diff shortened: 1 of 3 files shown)


reticulatedpines:
Hi.  I don't want to tie the repo to Github, and building is currently very easy, so I don't think we want to use GitHub Actions.

I don't know what problem the make file change is supposed to fix - can you explain this?  Thanks

petabyt:
> I don't know what problem the make file change is supposed to fix - can you explain this? Thanks

Clearly it doesn't fix anything, it adds a target to compress all camera release zips into a single archive (not sure why it needs to do that)

PR #202, opened 2025-06-25
modules: sd_uhs

radosek:
Port from `bilalfakhouri/magiclantern_hg_02/tree/crop_rec_4k_mlv_snd` with replacement for `patch_instruction`. Included in EOS-M build.

### DIFF
--- a/modules/Makefile
+++ b/modules/Makefile
@@ -52,7 +52,8 @@ ML_MODULES = \
     adv_int \
     crop_rec \
     dev_tools/edmac \
-    yolo
+    yolo \
+    sd_uhs
 
 .PHONY: clean $(BUILD_DIR)
 
--- a/modules/sd_uhs/Makefile
+++ b/modules/sd_uhs/Makefile
@@ -1,0 +1,8 @@
+# define the module name - make sure name is max 8 characters
+MODULE_NAME = sd_uhs
+MODULE_OBJS = $(BUILD_DIR)/sd_uhs.o
+TOP_DIR = ../..
+
+# include modules environment
+include $(TOP_DIR)/modules/Makefile
+.DEFAULT_GOAL := $(BUILD_DIR)/module_complete
--- a/modules/sd_uhs/README.rst
+++ b/modules/sd_uhs/README.rst
@@ -1,0 +1,9 @@
+SD overclocking
+===============
+
+Playing Russian Roulette with your data.
+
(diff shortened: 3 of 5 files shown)


reticulatedpines:
Would be good to get a working version into the repo.  What testing has been done?

radosek:
@reticulatedpines we spoke about it in `#general-development-discussion` on Discord. I tested only EOS-M, and the benchmark was fine. Somebody would have to test it on other cameras and maybe do a few more tests with actual speed during recording, as I got those speeds below.

```
| Resolution | Duration   | Speed        |
|------------|------------|--------------|
| 1736       | +3:20 min  | ~31.15 MB/s  |
| 2240       | +3:20 min  | ~64.72 MB/s  |
| 2520       | 6-23 s     | ~75.05 MB/s  |
```

reticulatedpines:
Okay, known to work on one model only, a little testing done on that model, speed improvements unknown.  Which is fine, everything has to start somewhere.

radosek:
Unfortunately, I have just one camera.

Only change was from `patch_instruction` to

```c
// Replacement for the legacy patch_instruction().
static inline int patch_instruction32(uintptr_t addr,
                                      uint32_t old_val,
                                      uint32_t new_val,
                                      const char *desc)
{
    struct patch p = {
        .addr = (uint8_t *)addr,
        .old_value = old_val,
        .new_value = new_val,
        .size = 4,
        .description = desc,
        .is_instruction = 1,
    };
    return apply_patches(&p, 1);
}
```

So I suspect that it should work on all cameras as before. One of the places where it was used (`patch_instruction`) was for EOSM.

Bench speed shows as like this for Lexar Silver Plus (MicroSD), but as per discoussion on Discord, if we use use old `card_bench.c` the speeds are same as before (around 5 MB/s less).

image

reticulatedpines:
Yup, understood.  But here there is no "as before" - this module has never been part of this repo.  It will need testing on a range of cams.  I'm not asking you to do that, just letting you know it needs to happen before I can take this code.

reticulatedpines:
Have started work on this now.  Rebase was easy, and module works on 70D.

I'm going to change the code a lot because the current approach is hard to adapt to modern cams.

reticulatedpines:
I changed some things and fixed some bugs, this is fixed by commit 9c1cd0a3c910313f30d0dbdfe5a4035cf3c32538 and some earlier ones.

PR #203, opened 2025-06-28
Fix deprecation warning

anibarro:
Replace the [deprecated distutils](https://mail.python.org/archives/list/python-dev@python.org/thread/TXU6TVOMBLQU3SV57DMMOA5Y2E67AW7P/) call in Python 3.10-11, by [shutil](https://docs.python.org/3/library/shutil.html#shutil.which) ([recommended replace](https://peps.python.org/pep-0632/#:~:text=distutils.spawn.find_executable%20%E2%80%94%20use%20the%20shutil.which%20function))

### DIFF
--- a/modules/readme2modulestrings.py
+++ b/modules/readme2modulestrings.py
@@ -61,9 +61,8 @@ def declare_string_section():
     
 def is_command_available(name):
     """Check if command `name` is on PATH."""
-    import distutils.spawn
-    from distutils.spawn import find_executable
-    return find_executable(name) is not None
+    import shutil
+    return shutil.which(name) is not None
 
 # return the first command of the list that can be found on the OS
 def get_command_of(commands):


reticulatedpines:
Hi - what kind of testing has been done for this change?  In particular, distutils / spawn stuff is often a pain on different OSes.

anibarro:
This particular call is just about checking if a command exist on path. I added the documentation if you want to read more, but distutils was already removed in Python 3.12.
I tested it on Windows, WSL2 Ubuntu 22.04 distro.

reticulatedpines:
Yes, I understand what this code does, and I've had direct experience of seemingly trivial changes in spawn / distutil area causing real bugs on different OSes.  Since not breaking the build system is very important, I try to be cautious when changing it.

I am using python 3.13 and it has distutils.  Perhaps that is a Debian quirk.

I've been aware of this warning for a long time, and it causes no problems that I'm aware of.  I want to fix it, simply because it's a little noisy in the build output, but I'd prefer fixing it by moving the import out of the function - that always felt weird to me.

This is a small change, and one I think can be done a better way.  I intend to close this PR unless you have a compelling reason why this approach is both needed now, and must be done in this manner.

anibarro:
All good for me, fix it the way suits you better

PR #208, opened 2025-07-27
make_bootable: rewrite for SCRIPT flag and allow for flags removal

kitor:
~~Requires testing. I've done only limited one with FAT16 card so far.~~ Tested on all types of cards successfully.

Rewrite of make_bootable.py to allow:
- setting each flag individually
- optionally erasing unset flag (eg convert bootable card to non-bootable or script only)

After a discussion, updated so it defaults to similar-to-previously behaviour (ran with just device path will set all flags on)

From a discussion, scripts are also renamed, got rid of old .sh script and pointed docs to the new one.


tools/make-bootable/make_bootable.py
```diff
@@ -1,49 +1,107 @@
...
+            print(f'  removing BOOTDISK at offset {off:x} (Boot code)')
+            dev[off:off+8] =  b'\0\0\0\0\0\0\0\0'
+        if offsets.script and not args.script:
+            off = base + offsets.script
+            print(f'  removing SCRIPT at offset {off:x}')
+            dev[off:off+6] = b'\0\0\0\0\0\0'
+
+
+def setExfat(device, base, args):
+    print(f'Updating EXFAT VBR at {offset:x}')
```
petabyt:
```
Identified EXFAT partition
Traceback (most recent call last):
  File "/home/daniel/Documents/mlinstall/make_bootable.py", line 165, in <module>
    main()
    ~~~~^^
  File "/home/daniel/Documents/mlinstall/make_bootable.py", line 153, in main
    setExfat(device, 0, args)
    ~~~~~~~~^^^^^^^^^^^^^^^^^
  File "/home/daniel/Documents/mlinstall/make_bootable.py", line 104, in setExfat
    print(f'Updating EXFAT VBR at {offset:x}')
                                   ^^^^^^
NameError: name 'offset' is not defined
```

kitor: Fix'd in last commit.

PR #210, opened 2025-08-09
We often don't declare function params in header files, we should

petabyt:
Perhaps this was intentional but it's not necessary.

EDIT: original submission had incorrect title and no description of the problem here.  The reported problem was that we have incompatible double definitions of some functions, e.g. lens_wait_readytotakepic() in lens.h.  The definition was not incompatible, but it shows a real problem, which is larger.  Probably it is not serious, but, it is ugly.

### DIFF
--- a/src/lens.h
+++ b/src/lens.h
@@ -895,7 +895,6 @@ void set_htp(int value);
 
 /* camera ready to take a picture or change shooting settings? */
 int job_state_ready_to_take_pic();
-void lens_wait_readytotakepic();
 
 /* force an update of PROP_LV_LENS outside LiveView */
 void _prop_lv_lens_request_update();


petabyt:
Same thing in ../../src/menu.c:143:12

reticulatedpines:
Contrary to the title (I'll change it), it's not an incompatible definition.  But it is weird: https://stackoverflow.com/questions/5929711/c-function-with-no-parameters-behavior/5929736

Omitting the params leaves the number and type of params undefined.  Which apparently is compatible with a prior definition that does define them.

You can easily find all the occurrences where we don't define the params by by adding `-Wstrict-prototypes` to Makefile.globals, and running make with `FATAL_WARNINGS=y`.  A bit of grep and uniq magic suggests there are several hundred.

reticulatedpines:
Raised https://github.com/reticulatedpines/magiclantern_simplified/issues/212 for the very tedious looking audit work.  I expect this will find multiple categories of required fixes (hopefully most just need the single, well defined param declaration copied into the header).  I suggest each category gets a PR, when we know what they are.

petabyt:
Doesn't `-Wstrict-prototypes` also warn about a missing void from a parameter list? Do you want that?  
EDIT: Lol, that's the only thing that flag does.

I was only concerned about repeat definitions (which could potentially be confusing), and with the way I scanned the codebase I think I only found a few of those.

reticulatedpines:
Both the examples you gave are found by -Wstrict-prototypes.  _Conflicting_ declarations (you mean declaration, not definition, right?) are already a warning, and I already fixed all warnings.  I assume the reason these weren't found is because they don't conflict.

The non-conflicting repeated declarations are safe to remove.  But the other ones may not be, and we use that pattern a lot.  It seems best to fix them all, so we can enable the warning flag permanently.

petabyt:
I'm *not* talking about all the functions that are brought up by adding `-Wstrict-prototypes`. I think requiring strict prototypes is probably good but would require a lot of work to fix the warnings. The specific issue I raised here is about repeated declarations (sorry, not definitions) in the same file that are different.

The two examples I gave were found when I ran the codebase through clang. For example:
```
../../src/lens.h:898:6: warning: a function declaration without a prototype is deprecated in all versions of C and is treated as a zero-parameter prototype in C23, conflicting with a previous declaration [-Wdeprecated-non-prototype]
  898 | void lens_wait_readytotakepic();
      |      ^
../../src/lens.h:560:6: note: conflicting prototype is here
  560 | void lens_wait_readytotakepic(int wait);
```

petabyt:
Also, see mvr_rec_start_shoot. wtf.

reticulatedpines:
Yes, I understand you weren't talking about that, but because you found the problem, we are given the opportunity to find the best fix - that's good!  If we just fixed these two examples, we might forget or miss the hundreds that are not found the same way.  Finding bugs is hard.  Spending the time to establish the root cause for a problem, and trying to find the broadest sensible fix, is worthwhile.

Note that the clang warning states this is only a conflict in C23 and later.  We're C99 (well, gnu99, close enough) and likely to remain that way forever.  Consequently the part of the clang warning that applies here is only the deprecated prototype style.  -Wstrict-prototypes feels the right way to warn about that in C99.

Since your proposed change doesn't functionally alter the code in any way, I feel it can wait until we fix the whole problem.

reticulatedpines:
> Also, see mvr_rec_start_shoot. wtf.

Please describe what you mean rather than making me guess.

petabyt:
> > Also, see mvr_rec_start_shoot. wtf.
> 
> Please describe what you mean rather than making me guess.

mvr_rec_start_shoot is defined with no parameters but is used with a single int parameter. Undefined behavior!
```
daniel@desktop:~/Documents/magiclantern_simplified/src$ grep -rn . -e mvr_rec_start_shoot
./shoot.c:2424:void mvr_rec_start_shoot(){}
./shoot.c:2426:void mvr_rec_start_shoot(int rec)
./lens.c:1359:    mvr_rec_start_shoot(buf[0]);
./shoot.h:47:void mvr_rec_start_shoot();
```

Those 3 examples I gave you are all that clang found in src/.

reticulatedpines:
Thanks.  That's not undefined behaviour.  This is explained in the Stack Overflow post I linked to.  And -Wstrict-prototypes finds many more.

petabyt:
The stackoverflow post describes the behavior in a single translation unit. Here we're dealing with two separate translation units, shoot.c and lens.c. The caller expects an "unspecified (but not variable) number of parameters" and the definition in shoot.c has one parameter. These are two different ABIs. Undefined behavior, no?

Here's a better post: https://stackoverflow.com/questions/41803937/func-vs-funcvoid-in-c99

reticulatedpines:
Which part are you claiming has undefined behaviour?  And in which build configuration?  You left out a rather important line from shoot.c:
```
2424 void mvr_rec_start_shoot(){}
2425 #else // 50D only above, everything else below:
2426 void mvr_rec_start_shoot(int rec)
```

Only one or the other is used as the definition.  And do 50D builds use the line in lens.c?

This kind of stuff is why I would greatly prefer fixing all ambiguity!  Stopping using empty param lists in definitions means I don't have to think about this at all.

petabyt:
Technically both are undefined behavior.

reticulatedpines:
"Both"?  Both what?

petabyt:
> "Both"? Both what?

Both definitions you just sent me. When they are called by this line: `mvr_rec_start_shoot(buf[0]);`

reticulatedpines:
Is `buf[0]` type int?  If so, that seems fine with at least the `void mvr_rec_start_shoot(int rec)` definition.

petabyt:
GCC is probably just assuming the type that function. But there isn't a guarantee `buf[0]` will actually make it to that implementation!

reticulatedpines:
I don't think that's true.  But you're the one claiming undefined behaviour, so I say it's up to you to demonstrate it :)  To me it looks like it might possibly be undefined in the 50D case, but not the other.  Notably, as per your SO link, shoot.h does not provide a function prototype (which explains why the declarations in your original report don't conflict; only one provides a prototype).

There is some academic interest here, but the practical approach feels obvious to me: remove the need to spend this much time trying to interpret the standard, by removing all empty function parameter lists.

petabyt:
Well, it says it right here:

> If the expression that denotes the called function has a type that does not include a prototype, [...] If the number of arguments does not equal the number of parameters, the behavior is undefined.

I agree, add `-Wstrict-prototypes` and fix all the warnings. You seem pretty opinionated on how you want to go about doing these things so I'll let you do that.

reticulatedpines:
Directly above the part you quoted:
```
void func1() { ... } can still be called with any number of arguments
```

And:
```
since func1 doesn't have a prototype, a conforming implementation is not required to produce any diagnostics
```

In the non-50D case, we have no prototype from the header, we define the function as taking one int param, and we call it with one int param.  I see no undefined behaviour here.

In the 50D case, we have no prototype from the header, we define the function as _having no named parameters_, but we are still permitted to call it with any number of arguments.  They merely have no names inside the function, so we cannot refer to them within the body.  Again, I see no undefined behaviour.  Here I am much less sure because the standard is more complex and confusing.

Is it a strong opinion to suggest removing ambiguity by stopping using a deprecated part of C?  What would be the weakly opinionated alternative?

petabyt:
Regardless of whether we agree it's undefined behavior we both agree it's bad, so let's move on.

petabyt:
Are there any firmware stubs declared without a prototype?

reticulatedpines:
> Regardless of whether we agree it's undefined behavior we both agree it's bad, so let's move on.

Yup - that's why I kept going, to demonstrate that the C standard is, at least, time consuming to interpret.  If you're wondering if something in your code is undefined or not, probably best to just make it more explicit.

> Are there any firmware stubs declared without a prototype?

Strictly speaking, stubs are not even C.  see `include/stub.h` - all stubs are is a set of externally visible symbol names, with an associated value.  Sounds like you're anticipating some problem if they are used without a prototype?  Off hand, I don't know if this is possible.  I know when I add a stub but forget to declare a prototype before use, it will fail to compile.  I don't know if there are cases where it's possible to compile.

We typically declare a prototype in some header related to the stub functionality, e.g. ConnectReadEDmac() in edmac.h.

petabyt:
Ya I know, I even stole stub.h for hacking Fuji.

Once the refactor is done what's the procedure for testing?

reticulatedpines:
Okay, so what's the intent re the question about stubs and prototypes?

Appropriate testing depends on what categories of problems are found.  I'd guess the biggest bucket is "provably no problem", because we don't actually use functions in weird ways that the lack of prototype allows.  For those, no special testing needed.

petabyt:
> Okay, so what's the intent re the question about stubs and prototypes?

Sorry, I mean to say stubs with a prototype defined without an parameter list. If there's any code that passes arguments to them then that could be something to watch out for.

reticulatedpines:
Okay, makes sense.  Dunno, would need to check the several hundred reported problem lines from -Wstrict-prototypes.  Which I will, but not right now.

Issue #221, opened 2025-08-31
[Feature request] [6D2] Focus box autohide feature. Clean HDMI / Clear Overlays


evgeniimv:
Hey there, 

First, thank you Magic Lantern team for your huge milestone on extending Nightly builds for bunch of new Canon cameras.
I personally an owner of 6D Mark II, and I was waiting for this release from the very moment I got this camera.

So, to the bis, I saw some discussion around [desired feature set for new cams](https://github.com/reticulatedpines/magiclantern_simplified/issues/155)
And the comment left by @WalterSchulz that it would be a top `wish` to have `Clean HDMI` on new cameras.

I fully understand that getting it done will take time, thus I will make my best to support you on this.

One hint is that Canon 6D Mark II already has an option to hide most of the overlays from the `LiveView` screen EXCEPT `Focus Box`.
So, essentially to achieve `Clean HDMI/Clear Overlays` feature we may only need to implement `Focus Box` hiding.

WalterSchulz:
Please mark it as "Feature Request"

evgeniimv:
Would be nice to have a feedback on pushed PR related to this issue :)

PR #223, opened 2025-08-31
issue-221 add FEATURE_LV_FOCUS_BOX_AUTOHIDE to features list in 6D2

evgeniimv:
This is my first interaction/contribution to the project. So I had to configure my environment to have files to compile and to build a working ZIP. 
The motivation to contribute to the project was the fact that Im the owner of 6D Mark II, and I want to have my LiveView with AF w/o unnecessary overlays on the screen. 

I created an issue to track activity on a development https://github.com/reticulatedpines/magiclantern_simplified/issues/221 It basically explains that it would be nice to have a feature to hide a Focus Box while using a LiveView mode. 
This will essentially help to achieve a Clean HDMI since the rest can be removed/cleared with INFO button (basic functionality).

Testing. 
After compilation/packaging with `make -j6` (actually I had to use a bit updated command because Im not allowed to modify `/usr/bin` on a corp laptop)
```
make -j6 CC=/opt/homebrew/bin/arm-none-eabi-gcc AS=/opt/homebrew/bin/arm-none-eabi-as AR=/opt/homebrew/bin/arm-none-eabi-ar OBJCOPY=/opt/homebrew/bin/arm-none-eabi-objcopy LD=/opt/homebrew/bin/arm-none-eabi-ld READELF=arm-none-eabi-readelf STRIP=/opt/homebrew/bin/arm-none-eabi-strip
```
With slight modifications for `python3` and `gstat` I managed to have it packaged w/o errors.
image

Then with files generated in `build/zip` I reinstalled build on Canon 6D2 and verified the functionality. 

image
image
image

https://github.com/user-attachments/assets/495d9512-58f0-4805-87cb-1d6c1fab5082

On the top header I see the error message `EFLensComTask: stack warning: free=232 used=792` which seems to be flaky appearing from time to time.
I guess there could be a potential thread(s) optimisation needed. I will give more updates if it causes issues on longer use.


evgeniimv:
Update, atually Focus Box remains on the screen, when used with HDMI output. This feature, only clears Focus Box from the Canon display. 
Basically means we need full fledged Clear HDMI feature.
Sorry for a confusion. I will continue looking into this

reticulatedpines:
Yup, the newer cams draw in a different way, and it means our old code doesn't work.

If you trace what FEATURE_LV_FOCUS_BOX_AUTOHIDE enables, the work is done in clear_lv_afframe(), tweaks.c

You may want to join our discord, to ask questions about the differing graphics layers.  It's a lot to go into on here.

reticulatedpines:
For the stack warning, that's our code, see tskmon.c.  We have a few exceptions already, where stock tasks are very close to the max stack limit.  I don't see the one you mention, so I'd like to understand what triggers it before adding another exception.

```
195         #if defined(CONFIG_200D) || defined(CONFIG_6D2)
196         // this might be CONFIG_DIGIC_VII
197         if (streq(task_name, "RTCMgr") && free > 128)
198             return; // RTCMgr uses 796 of 1024, 228 free
199         if (streq(task_name, "idle") && free > 64)
200             return; // in Play mode, uses 128 of 256, 128 free
201         #endif
```

PR #227, opened 2025-09-15
Support for a Dev Container

michaelgregorius:
Add support for a Dev Container which is based on Arch Linux.

Please refer to the added README.md for more details.

Remove the `.*` line from the `.gitignore` so that the `.devcontainer` directory can be added.

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,0 +1,14 @@
+FROM archlinux:latest
+
+RUN pacman --noconfirm -Syu && \
+    pacman --noconfirm -S make python3 python-pip python-distutils-extra python-docutils gcc arm-none-eabi-gcc arm-none-eabi-newlib inetutils zip && \
+    rm /var/cache/pacman/pkg/*
+
+# Create the build user
+RUN useradd -m build && \
+    # Delete the build users password
+    passwd -d build && \
+    # Allow the build user passwordless sudo
+    printf 'build ALL=(ALL) ALL\n' | tee -a /etc/sudoers
+
+USER build
--- a/.devcontainer/README.md
+++ b/.devcontainer/README.md
@@ -1,0 +1,21 @@
+# What is this?
+This directory defines a "Dev Containers" environment that provides all necessary tools to compile "Magic Lantern" while keeping the host system clean of these dependencies.
+
+# Prerequisites
+* Install the "Dev Containers" extension for Visual Studio Code (`ms-vscode-remote.remote-containers`)
+* Make sure that Docker is installed and running
+
+# Usage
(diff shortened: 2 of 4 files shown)


reticulatedpines:
Hi, I don't see docker as being part of the main repo.  We already support native builds on Linux, Mac and WSL.

There is a separate repo that allows building via docker: https://github.com/reticulatedpines/magiclantern_docker_builder

This is intended for building only, not dev work.

If you can explain why you want this, we can work out what the best option is.

michaelgregorius:
Hi @reticulatedpines,

to me a [dev containers](https://code.visualstudio.com/docs/devcontainers/containers) is a tool to lower the barriers for potential contributors because they don't need to install all the necessary development packages onto their main system. The dev container provides a certain isolation from the host system while providing all the packages necessary to build and develop.

Due to using Docker they will also work on all different Linux distributions, i.e. contributors do not have to hunt for the development packages that they need for their specific distribution. They are just available through the container.

In case you want to try it out I propose the following:
* Install a portable version of Visual Studio Code, i.e. download the zip version and create a `data` folder next to the `bin` folder.
* Make sure that Docker is installed.
* Install the VS Code extension as described in the README.md
* Open the branch from this PR

Because usage of the dev container is optional it will also not create a hard dependency on Docker ("I don't see docker as being part of the main repo.").

Some investigation seems to point to the possibility of having the dev container definition in a separate repository. However, it's then much more cumbersome to use them. If it's in the main code repository developer just need to have Docker and the "Dev Container" installed and then everything will work "auto-magically".

reticulatedpines:
Thanks, that helps me understand your intention.  I wasn't aware of this integration between VSC and containers.  I've never used VSC and don't like using containers as a dev env (very useful for build environments).

We currently have no devs that use docker or Visual Studio Code.  Getting a working dev environment is already low friction.  When we get people interested in doing dev work, we get their dev environment up and running quickly.  The difficult parts are typically teaching people reverse engineering, along with information about camera internals.

If this was added to the repo, it would be a maintenance cost.  We are a very small team, and nobody on the team would be using it, making it awkward to maintain.

Having this as an option that I can point people at, should we ever get any devs that use VSC, could be useful.  I'd happily recommend people try it in that circumstance.

michaelgregorius:
@reticulatedpines, if you want we can keep this PR open and if people want to try it out, they can use it as a basis to give it a go. What do you think?

reticulatedpines:
I don't want PRs that are kept open forever (there's already a few draft PRs that I need to get people to tidy up).  Of course you are welcome to keep it open in a fork.

michaelgregorius:
Okay, closing this PR then. The branch in my fork is [dev_container_support](https://github.com/michaelgregorius/magiclantern_simplified/tree/dev_container_support).

Not sure how long that fork will live though because I am not actively working on magic lantern and this branch was rather something that was interesting to me and that I wanted to offer to the project. Can totally understand the concerns though.

PR #233, opened 2025-10-18
Refactor dump_fir.py for Python 3 compatibility

lclevy:
Updated to Python 3 compatibility and changed version to 0.6.

### DIFF
--- a/tools/indy/dump_fir.py
+++ b/tools/indy/dump_fir.py
@@ -1,13 +1,15 @@
 #dump decrypted thing
+#2025: Lorenzo3324@discord, python3 compatible
 
-from struct import unpack
+from struct import unpack, Struct
 import os
 import sys
 import array
 import ctypes
 from optparse import OptionParser
 
-version_id = "0.5 (22dec2012)"
+#2025
+version_id = "0.6 (oct2025)"
 
 def getLongLE(d, a):
    return unpack('<L',(d)[a:a+4])[0]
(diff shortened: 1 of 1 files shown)


reticulatedpines:
We don't use this file anywhere, and I don't know how to review the changes for correctness.

What's the motivation for this PR?

lclevy:
Hi, 
in this [talk](https://github.com/lclevy/d8_oracle/tree/main/hacklu2025_slides)
 I gave this afternoon, I explain how some Digic8 camera updates (Fir files) can be decrypted using emulation of a firmware dump. This script is an update of existing ML script. Dump_fir.py allows to display contents of decrypted firmware.

reticulatedpines:
I know what the script does.  I would like to understand *why* you want to update it, given it is not used.

lclevy:
Because, I updated it on my side, and I think it is normal to contribute back.

reticulatedpines:
In that case, thank you for the offer, but the changes are not needed here.

lclevy:
Python 3 compatibility is not needed? What are you afraid of ?

reticulatedpines:
No changes are needed to this file, we never use this file.  Taking changes means testing those changes, which takes time and effort.  And this file is so old, I don't even know what the original is supposed to do, so what tests would prove the changes are good?

You got some useful base code to work from, which presumably saved you time.  That's good!  But Magic Lantern no longer uses this code.  Why should I spend time reviewing and testing it, when it will continue to not be used?

lclevy:
For people reproducing experiments described on my talk

reticulatedpines:
That seems like a good reason to make the source available, but not a good reason why it wants to live in this repo, or why I should spend a few hours making sure the changes are good.

lclevy:
Available here too: https://github.com/lclevy/d8_oracle/blob/main/ml_scripts/dump_fir.py

PR #235, opened 2025-10-22
python3 comptibility and some code refactoring

lclevy:


reticulatedpines:
We don't actively use this tool.  It would be easier to take the changes if you made two commits, one purely for python3 compatibility, with any other changes split out.

PR #246, opened 2025-12-06
Wrap documentation to 80 columns (and grammar)

mechite:
Markdown is a great format because it allows us to write plain-text with
embedded formatting, and still also be able to wrap to the 80-column terminal
standard.

I made these changes because I find it more readable when you are not using
a markdown editor. This happens often enough for myself.

I made the choice to not wrap markdown links (`()[]`), as it does not affect
readability.

Changes:
- wrap to 80col
- lowercase and add missing periods to stuff like "e.g.", "etc."
- correct casing of "GitHub" (previously "Github")
- add "\*nix" to the message saying that BSDs may work.

  I also want to say that I did test building on OpenBSD (due to it being my OS
  of choice on my laptop) and it appears to work pretty well.
- remove the `<div style="page-break-after: always; visibility: hidden"></div>`
  that seems to be in some markdown files, not sure what it is for?

Reading the documentation in the GitHub web viewer, other than the spelling
corrections, you should notice no difference than the original files.

The point — now both representations are nice to look at.

These changes are opinionated, feel free to close if you disagree with them.
I suggest to squash merge (commit messages are blank here).

Signed-off-by: Mahied Maruf <contact@mechite.com>

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,135 +1,198 @@
 Introduction to Magic Lantern
 =============================
 
-Hello, potential camera hacker!  Have a Canon EOS camera and want to make it do extra things?  Wondering why Magic Lantern isn't on your camera yet?  Do you enjoy low-level messing around and exploring undocumented systems?  Or, you think you would and want to learn?  Give it a try, it's fun, if a little tricky sometimes.
+Hello, potential camera hacker!
+
+Have a Canon EOS camera and want to make it do extra things? Wondering why
+Magic Lantern isn't on your camera yet?  Do you enjoy low-level messing around
+and exploring undocumented systems? Or, you think you would and want to learn?
+Give it a try, it's fun, if a little tricky sometimes.
 
 Anyone who can help us keep things working and improving is greatly appreciated!
 
-Building ML should work on Linux, Windows (WSL), or Mac (probably BSDs generally).
+Building ML should work on Linux, Windows (WSL), or Mac (probably *nix, BSDs).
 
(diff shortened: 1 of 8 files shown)


reticulatedpines:
A bunch of notes in no particular order:

I don't want the 80 col change, it makes editing and review more hassle.  The diffs are noisier when change in one line necessitates moving words around in multiple other lines to keep them all below 80 cols.

Semantic changes should be separated from style changes.  As is it's hard to find the changes to text because every line is listed as changed due to 80 col formatting.

The div styling is required; when generate_guide.py is used, this makes parts work correctly in the HTML output.  Markdown is allowed to contain arbitrary HTML.

The normal rule for abbreviations is to capitalise the first letter, if the word that is abbreviated would normally be capitalised.  E.g, "e.g.", should it be at the start of a sentence.

We don't use traditional merges, we only ff merge dev.  I think that makes squash merges lose all history?  If so, that's probably a bad idea.

mechite:
> this makes parts work correctly in the HTML output

Right. I would have expected something like this.
I was running with the assumption it was from some markdown writing tool.

> I don't want the 80 col change, it makes editing and review more hassle.

Personally I disagree, I don't find it a common occurrence that edits are made in this way, but this is a matter of opinion.
Yours should take precedence.

> The normal rule for abbreviations is to capitalise the first letter

I don't see this commonly, but it's because I don't start a sentence with them.

---

The "rich diff" in GitHub makes the changes easier to see.
I did also notice some mistakes I made this way.

Closing because most of the changes are unwanted, and I don't want to waste your time.

PR #253, opened 2026-01-01
Add support for devbox

mogorman:
[devbox](https://www.jetify.com/devbox) is an environment management tool. it is similar to how you use docker but much simpler in execution. It provides all the tools needed to build magic lantern and locked to the same versions so everyone builds with same tools.

Not sure if this is wanted but it does make development a lot easier for me so I thought I would share.

### DIFF
--- a/Makefile.globals
+++ b/Makefile.globals
@@ -13,7 +13,11 @@ ifndef TOP_DIR
     endif
 endif
 
-ARM_BINPATH = /usr/bin
+ifdef ARM_BINPATH
+  ARM_BINPATH := $(ARM_BINPATH)
+else
+  ARM_BINPATH := /usr/bin
+endif
 ARM_ABI = none-eabi
 CROSS_COMPILE = $(ARM_BINPATH)/arm-$(ARM_ABI)-
 HOST_CC = gcc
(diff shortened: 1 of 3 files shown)


reticulatedpines:
It's difficult to take on a practical level: I don't think anyone in the current dev team uses Nix, or has any experience with reviewing or maintaining Nix code.

More generally, I don't like tying the code to any specific dev env.  Having devs on different platforms and environments has so far been a useful source for finding bugs in the build system.

Are there specific problems that this helps you with?  Perhaps there is some other way the existing build system can be improved.

mogorman:
this just made it really easy for me to build the project in same way docker would. i use devbox tool to manage the different projects i have to build and keep my machine clean, the tools it pulls in are isolated. so if a different project i build uses a different version of the cross compiler i have no conflicts.  I understand if its something you wouldnt be interested in but thought id share.

reticulatedpines:
Okay, thanks.  That sounds like a useful thing for you, but not something we need.  What are you looking to do with ML generally?  Are you in our Discord?

mogorman:
no worries. im in discord as mog, im a canon 70d user so looking to improve that for making movies.

PR #268, opened 2026-05-10
70D: Build system fixes (CONFIG_QEMU, dead code cleanup)

peva3:
I fixed the CONFIG_QEMU default. It was set to `y` for all builds which meant every build was a QEMU build. The QEMU testing code is fine in emulation but causes a red LED on the physical 70D. I changed it to `?= n` so `make` produces a hardware build and `make CONFIG_QEMU=y` produces a QEMU build.

I cleaned up 14 dead `#if 0` blocks in `src/module.c`, `src/arm-mcr.h`, and a few other files. These were historical TCC compiler structs, redundant typedefs, and abandoned GDB stubs. Nothing functional was touched.

I fixed `test_qemu.sh` so it handles different firmware versions instead of being hardcoded to .202. I also removed non-existent QEMU model names from the case statement (EOSM3, EOSM5, EOSM10).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
+syntax: glob
 *.gpr
 *.gar
 *.qcow2
@@ -32,7 +33,6 @@ tags
 *.i
 *.log
 *.pyc
-__pycache__/
 build_tools/xor_chk
 docs/html
 docs/latex
@@ -77,43 +77,4 @@ modules/*/*/build
 developer_guide/developer_guide.html
 developer_guide/developer_guide.pdf
 
-# qemu-eos build artifacts
-qemu-eos/build/
-qemu-eos/*.o
-qemu-eos/*.d
-qemu-eos/config-all-devices.mak
-qemu-eos/config-all-disks.mak
-qemu-eos/config-host.mak
-qemu-eos/config-status.mak
-qemu-eos/config-targets.mak
-qemu-eos/Makefile.ninja
-qemu-eos/build.ninja
-qemu-eos/stubs/
-qemu-eos/trace/trace-generated.c
-qemu-eos/trace/trace-generated.h
-qemu-eos/trace/trace-generated-timestamp.c
-core.*
-qemu-eos/magiclantern/core.*
-qemu-eos/magiclantern/disk_images/*.qcow2
-qemu-eos/ui/keycodemapdb/
-qemu-eos/dtc/
-qemu-eos/capstone/
-qemu-eos/slirp/
-
-# ROM dumps - only track essential ones (no RAM dumps, no placeholders)
-roms/*/RAM*.BIN
-roms/*/placeholder*
-roms/**/ROM*.BIN.gz
-
-70d-latest/
-70d-dev/
-
-# Build logs
-logs/
-
-# Release archives
-magiclantern_70D-*.tar.gz
-
-70D.gpr
-70D.rep/
-70D_IMPORT.txt
+syntax: regexp
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,3 @@
-syntax: glob
 *.gpr
 *.gar
 *.qcow2
@@ -33,6 +32,7 @@ tags
 *.i
 *.log
 *.pyc
+__pycache__/
 build_tools/xor_chk
 docs/html
 docs/latex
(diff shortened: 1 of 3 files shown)


reticulatedpines:
I'll do a first pass quick review here, rather than in the commit.  It's going to sound quite harsh, sorry!  I really recommend you ask us a few questions before making a big PR.  Here you are doing several things that we explicitly don't want, and have discussed previously, but you didn't know that.

As a first overall point, this shouldn't be one commit.  E.g. there are changes for 70D, but also changes to qemu that are nothing to do with 70D.  Those should go in separate commits.  If a PR has a theme, a commit should do one thing, well, not several unrelated things.

- CONFIG_QEMU does not default to y.  No builds default to this setting, and builds with CONFIG_QEMU set won't boot on phys cams (and builds work fine on phys cams here).  Can you explain why you believe it defaults to y?
- I don't want deploy targets as part of the make system (it's complicated to do reliably, would need testing on multiple OS, often needs to run as root, etc).  It looks to be copying to a dir that doesn't exist, so this will cause everyone else's builds to fail
- I don't want the #if 0 blocks removed.  This is commit noise for no gain at best, and in some cases removes comments that are marked as desired
- `@rm -f $(ZIP_MODULES_DIR)/*.mo $(ZIP_MODULES_DIR)/*.sym 2>/dev/null`  - this is dangerous and can delete files inappropriately, since `--` is not used to terminate arguments.  The intent is unclear to me so I don't know why you want this change
- You remove null pointer checks *that are required* on some cams, definitely can't take that
- Certainly don't want all the qemu-eos stuff added to gitignore.  Qemu-eos is not part of the main ML repo, and doesn't want to be
- Certainly don't want ignores related to rom files.  Never mix rom files into the repo dir structure, and don't make it easy to accidentally do this!
- In general, never change how the build system functions without consulting with the existing dev team, and performing appropriate testing.  We have to be cautious about bad builds, as we risk physical damage to other people's hardware if we make mistakes
- There are already scripts for running qemu with a given model cam, primarily run_qemu.py.  I don't see the utility in another one as test_qemu.sh (and shell is harder to maintain).  Probably this is not desired, but I don't know your reasoning.

peva3:
Thank you for the first pass and I appreciate it. I went back and realized I was actually on nikfreak's fork originally and that's why some of the config files aren't the same as mainline... I'm actually not sure how I did that, I was looking at a lot of people's changes from the 70d thread on the ML forums and I think I grabbed it there.

Heard on the other points.

peva3:
Updated this PR based on your feedback. I removed:
- The CONFIG_QEMU override (you're right -- the upstream doesn't have this issue, it was specific to the fork I was working from)
- The deploy target
- The test_qemu.sh script
- The .gitignore additions
- The #if 0 block removals

This is now just the Makefile fix for firmware version handling and the dead file cleanup. Let me know if there's anything else to adjust.

peva3:
Also removed the .gitignore additions (QEMU/ROM ignores) per your feedback. The PR now only has the firmware version fix in platform/Makefile.

reticulatedpines:
Re the gitignore change, when you have a mistake in code not yet in the repo, it's better to never put it in, rather than put it in then take it out again.  The latter adds two commits and means the intermediate commit is probably broken, which is annoying for e.g. git bisect.

I also noticed this time that you have modules in modules.include that don't exist in the repo.  This will cause builds to fail when the copy doesn't work (they probably work locally for you because you didn't clean your modules dir).

peva3:
What's the best avenue for discussion form your comment on the closed PR? The forums, discord, etc. just let me know and I'll move my responses there

reticulatedpines:
Discord is best for quick discussion: https://discord.gg/vJCWmmRJhb

reticulatedpines:
Hi @peva3 - any update here?  I haven't seen you in Discord.  I've been holding off doing more reviews until I know what you're doing.

peva3:
Both my work projects are in a sprint to submission for the next few weeks so I won't get a lot of time to work on this unfortunately. 

I've made changes on my own repo based of the feedback and want to do more hardware testing for the modules I'm working on. If you're fine with me making updates to the PRs I'm happy to do that, but I'll ping you on discord before I do anything.

reticulatedpines:
That's okay, there's no hurry.  Sounds like I should hold off spending the review time.

One thing I found time to test because it surprised me, was the claim that `call("TurnOnDisplay")` triggers a crash.  I can confirm this works fine on my 70D, as does TurnOffDisplay.  So I feel something else is going on with your code.

reticulatedpines:
Hi again @peva3 - will you be able to make any progress cleaning up and splitting these PRs?

peva3:
Hey, sorry, I've moved on to other stuff, feel free to close them all out.

reticulatedpines:
Okay, thanks.

PR #269, opened 2026-05-10
70D: Enable CONFIG_AUDIO_CONTROLS + 96kHz sample rate

peva3:
I enabled CONFIG_AUDIO_CONTROLS for the 70D. This is the first DIGIC V camera with ML audio controls working. The flag unlocks 8 features: analog gain (0-32dB), digital gain, AGC toggle, input source selection, wind filter, mic power, and headphone monitoring.

I added a sounddev_active_in NSTUB at 0xFF11936C. The my_sounddev_task override uses this during init to activate the audio device. The address is the same as SoundDevActiveIn, but the snake_case version is what the audio code calls.

I added a null guard in audio-ak.c. On the 70D, HOTPLUG_VIDEO_OUT_STATUS_ADDR is 0 because there is no HDMI hotplug detection. Without the guard, enabling CONFIG_AUDIO_CONTROLS would dereference null.

I added 96kHz to the mlv_snd sample rate selector. The AK4646 codec supports this rate. It is just one entry in the existing array.

The task override at audio-ak.c:643 replaces Canon's sounddev_task with ML's version. It sleeps 1500ms at startup, creates fake semaphores to disable Canon's ALC loop, then writes 10 I2C register values to the AK4646 through _audio_ic_write. After init it polls a semaphore every 100ms to re-apply settings.

### DIFF
--- a/modules/raw_video/mlv_snd/mlv_snd.c
+++ b/modules/raw_video/mlv_snd/mlv_snd.c
@@ -81,8 +81,8 @@ static volatile uint32_t mlv_snd_frames_queued = 0;
 static volatile uint32_t mlv_snd_frames_saved = 0;
 
 
-static uint32_t mlv_snd_rates[] = { 48000, 44100, 22050, 11025, 8000 };
-#define MLV_SND_RATE_TEXT "48kHz", "44.1kHz", "22kHz", "11kHz", "8kHz"
+static uint32_t mlv_snd_rates[] = { 96000, 48000, 44100, 22050, 11025, 8000 };
+#define MLV_SND_RATE_TEXT "96kHz", "48kHz", "44.1kHz", "22kHz", "11kHz", "8kHz"
 
 static uint32_t mlv_snd_in_channels = 2;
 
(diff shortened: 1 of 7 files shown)


reticulatedpines:
This one is a more sensible size, which helps :)  It still suffers from unrelated changes being in a single commit.  If the audio controls stuff works reliably, that would be a good upgrade.

If you want to change whitespacing, always do that in its own commit - then we can diff in ignore whitespace mode and the review becomes much easier.  While there are exceptions, in general, don't bother making whitespace changes, it's a waste of your time and ours.

The 96kHz mode is interesting, but I would have to test it first before taking the change.  If this exists on other cams (likely if it's on 70D), it would be nicer to make this a FEATURE flag and expose it via that, not cam specific checks.

I don't understand why you want to add sounddev_active_in, when you already have a symbol defined for this function.

The null pointer change is inappropriate without testing on other cams.  Some cams *deliberately* map data in the first 4kB.  All Digic 4 and 5 cams *can* read and write through a 0 pointer.  Sometimes they do this and the OS relies on it working.  Thus, your guard may help 70D but break other cams.

Don't remove `#if 0` stuff.  It's a waste of time at best, and sometimes removes useful comments.

Why have you changed how module loading works?  There is no explanation.  Perhaps this is a mistake, since you call `module_load_all_for_qemu()`, inside a CONFIG_QEMU guard, but no such function exists.

We can't accept a change to init.c that changes the boot process for all cams, without testing on all cams.  The comment suggests this is something to do with automated testing.  I would be happy to improve automated testing, but you will need to discuss your approach with the dev team before doing anything in this area.

The testing code in the CONFIG_QEMU guard might be easier to accept, but the comments are very confusing, and this feels a very inappropriate change in a PR for 70D audio support.  Since I have no idea what this file is for, what creates it etc, I would guess this is a mistake.

Likewise, the ptpip stuff and the comments about socket API probably shouldn't be in here?

WalterSchulz:
The audio features in Digic 4 cams do not have an 96 kHz mode.
And audio features from Digic 5 never made it to main. There was an attempt to make the audio chip work but ... 
And there is no 96 kHz option (Canon menu) in the Digic 5 cams I know.
-> AFAIK "96 kHz" on other cams doesn't exist.

PR #277, opened 2026-06-15
Update raw.c 80D DXO DR data

WalterSchulz:
- Deleted "//same sensor" which is most likely a copy-paste error.
- Replaced DR data with correct ones from DXO "screen". Previous entries were falsly transfered from "print" data.


reticulatedpines:
Closed by 436ae766f

PR #278, opened 2026-06-15
Update raw.c R5 DXO DR data

WalterSchulz:
- Deleted comment "// TODO: DxO graph is corrupted, so leaving R values for now", DXO data are backed up by Photons To Photos measurements. Hump in curve may be caused by amp gain manipulation (guess!)
- Transfered data frrom DXO DR "screen" graph.


reticulatedpines:
Closed by ea2d2a2aa

PR #281, opened 2026-06-19
EOS R (DIGIC 8): MMU ROM-patching + serial-flash RE [WIP]

jed2nd:
**Draft / WIP** — the integrated EOS R (DIGIC 8) research branch. This shares *everything* from the bring-up so the community has the full picture; the clean, mergeable feature subsets are already split into focused PRs (#279, #280). Not all of this branch is merge-ready — what's what is noted below.

## Headline: first DIGIC 8 R with working ROM patching (`CONFIG_MMU_REMAP`)
The R can now have ML patches applied to its (RAM-remapped) ROM — the foundation everything else on D8 needs. Getting there took two HW-confirmed fixes:
- **Red-LED-at-boot root cause:** `generic_mmu_space` carried an `__attribute__((aligned(0x10000)))` that forced a huge BSS linker hole, pushing ML's footprint past the R's tight `user_mem` budget → `my_create_init_task` aborts → solid red LED. It was never an mmu_init logic bug. Fix: drop the alignment + add a one-page slack (the L1 table is aligned internally).
- **`dcache_clean_multicore` HW crash:** Canon's inter-core cache broadcast (`0xE008E23E`) faults when called this early in `mmu_init` (inter-core HW not up). No-op'd for the R's cpu0-only remap (HW-bisected: with it = red LED, without = boots with full UI).

Confirmed on hardware: ML boots with full UI, and a ROM code-patch detour is live on cpu0 (SFDIAG dump).

## Reverse-engineering captured on this branch
- **Serial-flash driver fully mapped** — `ReadBlockSerialFlash` @0xE03C10C4, the driver struct @0x62bc (handle at +16), `InstallSerialFlash`/`UninstallSerialFlash`, deep-power-down. Canon installs the SF, reads the property DB at boot, then powers it down (handle→0). This RE is what produced the captured property data that lets **qemu-eos PR #18** boot the R past the property wall to `K424 READY`.
- **The R's exposure-property encodings** — shutter (hi-byte signed Tv·3), ISO (byte-1 code, 15=ISO100), aperture (byte-1 Av code, 9=f/2.8); all "value in a high byte, 3 units/stop". Clean capability-flagged feature in **#280**.
- **Clean capture path** via CameraConductor remote-release (no saving-hang) — **#279**.

## What's mergeable vs bring-up
- **Mergeable, already split out:** #279 (capture fix), #280 (shutter + ISO + aperture + HDR bracketing, capability-flagged).
- **Headline new, could be its own clean PR:** `CONFIG_MMU_REMAP` ROM patching (`src/patch_mmu.c`, `src/mmu_utils.c`, `platform/R.180/function_overrides.c`, stubs).
- **Bring-up / diagnostic, not for merge:** the serial-flash capture + active-read probes and the MPU/SF dump tooling in `src/init.c` / `src/debug.c` (the exposure-RE probes have already been pruned).

Happy to split `CONFIG_MMU_REMAP` into a focused standalone PR if that's preferred over this umbrella WIP.


reticulatedpines:
Very interesting, sounds like great work - will take me a while to review it!

The `dcache_multicore_clean` thing might not be the correct cause (but it's worth considering).  Can you patch rom0 without this change?  We found that rom1 is initialised later on this cam (even normal early reads fail), so it might be that instead.

If it *is* `dcache_multicore_clean` not working early, it's a bit unexpected (code is identical on digic 7 and 8, and it works on 7).  Certainly not impossible, we know other internal stuff has changed.

Aligned change I'll have to consider very carefully.  There are a bunch of alignment rules in ARM docs around MMU usage, and it's very important to meet them (hard fault and immediate hang if it fails to parse the tables).  Spotting the binary size going too large is good, we really should check that as part of the build process and fail / warn if so.  It's tedious work because it's a per cam size limit and involves messing with the build system.

More notes later :)

Issue #51, opened 2022-01-10
Refactor ptpcam (fix warnings / bugs)


coon42:
`contrib/ptpcam` gives alot of warnings when compiling and tends to crash when exiting with `ctrl+c` while it is running. We may refactor it a bit to make it more stable.

reticulatedpines:
Trying to clean up old PRs and issues.  Can you still repro this?  If so, can you give instructions on how?  Doesn't look like this is intended to build on Linux - Windows only?

coon42:
It's probably intended to build on Windows but it also works fine on Linux (ubuntu 18.04)

I think warnings got reduced by alot meanwhile. There are only some macro redefinitions left like:

```
"htobe16" redefined
"htobe32" redefined
"be16toh" redefined
"be32toh" redefined
```
etc.

Can be reproduced by executing `make` in `ptpcam` directory on `dev` branch using ubuntu 18.04

reticulatedpines:
Doesn't build for me, errors on missing <usb.h> - some dev library provides this I guess?  Could do with an improved README or something.

coon42:
yes, I think it should work after installing `libusb-dev`

reticulatedpines:
Good, builds for me with that added.  Still has a few compile warnings for me so I'm happy to leave this ticket for that part of the work.

"tends to crash" is not descriptive enough for a useful bug report, though I don't doubt you :)

Please file crashes you care about as separate issues, with repro instructions and stack traces / valgrind output etc.

coon42:
I agree that those are two different issues. We should only handle one issue per ticket, thus the `ctrl+c` crash issue has been moved to https://github.com/reticulatedpines/magiclantern_simplified/issues/65. I may try to fix the remaining duplicated definitions so this ticket can be closed.

reticulatedpines:
I think I fixed all of these when cleaning up the code before fixing some null ptr derefs.

Can you confirm?

PR #282, opened 2026-06-25
ptpcam: fix SIGINT NULL-deref crash before camera is connected

### LINKED ISSUE #65: ptpcam tends to crash when exitting with ctrl+c
As discussed in https://github.com/reticulatedpines/magiclantern_simplified/issues/51 `ptpcam` does not shut down properly but tends to crash when it is terminated via `ctrl+c`. It's not critical but it would be a nice fix.

chrisdebian:
Fixes #65.

## What's wrong

`ptpcam_siginthandler()` dereferences `globalparams->data` unconditionally. `globalparams` is only assigned inside `init_ptp_usb()`, which runs *after* a camera has been found and opened. `signal(SIGINT, ...)` is registered earlier in `main()`, before device discovery happens.

So there's a window — USB enumeration taking a moment, or simply running `ptpcam` with no camera attached — where Ctrl+C fires the handler while `globalparams` is still `NULL`, and it segfaults instead of exiting cleanly. This reproduces 100% of the time by just starting `ptpcam` and hitting Ctrl+C immediately, no camera required.

## Fix

Guard the `NULL` case and `exit()` immediately, since there's nothing to close yet. The existing best-effort cleanup path (`close_camera()`) for an already-connected camera is unchanged.

## Testing

I don't have a compatible Canon body to drive the full PTP/USB path, but the crash itself is a plain signal-handling bug independent of any camera, so I verified the logic with an isolated repro matching the same pointer-lifecycle pattern (global set late, dereferenced unconditionally in a `SIGINT` handler):

- Before the fix: `SIGINT` before the pointer is set → segfault (confirmed).
- After the fix: `SIGINT` before the pointer is set → clean exit, no crash.
- After the fix: `SIGINT` after the pointer is set (simulating a connected camera) → existing cleanup path still runs unchanged.

Scoped narrowly to the crash itself rather than a broader signal-safety rewrite of `close_camera()`'s libusb calls — that's a bigger change that probably belongs with #51 (general ptpcam refactor) rather than this fix.

### DIFF
--- a/tools/ptpcam/ptpcam.c
+++ b/tools/ptpcam/ptpcam.c
@@ -194,11 +194,19 @@ help()
 void
 ptpcam_siginthandler(int signum)
 {
-    PTP_USB* ptp_usb=(PTP_USB *)globalparams->data;
-    struct usb_device *dev=usb_device(ptp_usb->handle);
-
     if (signum==SIGINT)
     {
+	if (globalparams==NULL)
+	{
+	    /* No camera connected yet (e.g. still enumerating USB devices),
+	     * so there is nothing to close - exit immediately rather than
+	     * dereferencing a NULL globalparams below. */
+	    exit (-1);
+	}
+
+	PTP_USB* ptp_usb=(PTP_USB *)globalparams->data;
+	struct usb_device *dev=usb_device(ptp_usb->handle);
+
 	/* hey it's not that easy though... but at least we can try! */
 	printf("Got SIGINT, trying to clean up and close...\n");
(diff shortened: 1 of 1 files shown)


reticulatedpines:
Thanks for this, nice clean description of the problem.  I did a slightly wider fix, and cleaned up the whitespace, in the five commits from here onwards: https://github.com/reticulatedpines/magiclantern_simplified/commit/8bc122633a8bd7d945e3485930541c8c1d719fd3

Issue #286, opened 2026-07-25
Lua module fails to build with ARM gcc 15


reticulatedpines:
There is a complicated bug that I don't fully understand that prevents Lua module building with arm-none-eabi-gcc version 15 (and up, probably).

The failure is at link time due to a discrepancy in visible symbols.  The cause of this is to do with bad inclusion of system headers, while linking against static dietlibc.  One of the problems is related to inclusion of `<sys/cdefs.h>` - this looks like a system include but is sourced from a local dir, we're generally very sloppy about use of `""` vs `<>`.  We build with `-nostdlib`, but do not use `-nostdinc`.  System headers for GCC 15 (at least on Debian) change how some macros work, changing which header files make symbols visible.

Very probably we should avoid using system definitions for symbols at all, certainly we should stop using them when we provide them via our included static dietlibc.

With gcc 15, when building in `modules/lua`, the start of the error log looks like this:
```
In file included from /usr/arm-none-eabi/include/setjmp.h:42,
                 from lua/ldo.c:13:
/usr/arm-none-eabi/include/machine/setjmp.h:15:13: error: expected ';' before 'typedef'
   15 | _BEGIN_STD_C
```

coon42:
I was also unable to build on Arch linux and tried out claude code AI to find a fix. I can now build again by applying the commits of the following PRs:

https://github.com/reticulatedpines/magiclantern_simplified/pull/287
https://github.com/reticulatedpines/magiclantern_simplified/pull/288
https://github.com/reticulatedpines/magiclantern_simplified/pull/289
https://github.com/reticulatedpines/magiclantern_simplified/pull/290

However, we should not apply this blindly but verify this. I think the first 3 PRs are fine but I am not sure on the final one.

PR #287, opened 2026-08-01
modules: fix module build on Python 3.12+

coon42:

Here is the AIs report:

### What
  Replace `distutils.spawn.find_executable` with `shutil.which` in
  `modules/readme2modulestrings.py`.

  ### Why
  `distutils` was deprecated in Python 3.10 and removed from the stdlib in 3.12
  (PEP 632). On any distro shipping Python >= 3.12 the script dies before emitting
  anything:

  ModuleNotFoundError: No module named 'distutils'

  This looks like a docs problem but is not. `module_strings.h` is generated from
  each module's `README.rst` and compiled into the `.mo`, where it supplies the
  Name/Author/License/Summary text shown in the Modules menu — and
  `modules/Makefile:182` makes *every* module object depend on it. When the script
  fails, no module builds at all and no cam can produce `magiclantern.zip`.

  Debian trixie (Python 3.13) hits this out of the box; so does current Arch.

  ### Compatibility
  `shutil.which` is stdlib since Python 3.3. The script already required 3.3+ (it
  catches `FileNotFoundError`) and is python3-only (shebang, plus
  `PYTHON = python3` in `Makefile.globals`), so the minimum supported version is
  unchanged and there is no longer an upper bound. No f-strings or walrus
  operators anywhere in the script or the local modules it imports, so 3.3 remains
  the true floor.

  `shutil.which` is also the more correct test: `find_executable` does not check
  the execute bit on POSIX, so a non-executable `rst2html` would previously have
  been reported as found, then failed when the build tried to run it.

  Installing `setuptools` for its `distutils` shim would only defer the problem —
  that shim is itself deprecated and slated for removal.

  The script still needs `rst2html` on PATH (`python3-docutils` or your distro's
  equivalent); that dependency is unchanged.

  ### Testing
  Built `platform/200D.101` with `arm-none-eabi-gcc 15.2.1`: all 23 default
  modules and `magiclantern.zip`. Not tested on a physical camera — this only
  affects host-side build tooling and cannot change emitted code.

### DIFF
--- a/modules/readme2modulestrings.py
+++ b/modules/readme2modulestrings.py
@@ -4,6 +4,7 @@
 import sys, re
 import subprocess
 import os
+import shutil
 from datetime import datetime
 
 from align_string_proportional import word_wrap
@@ -61,9 +62,7 @@ def declare_string_section():
     
 def is_command_available(name):
     """Check if command `name` is on PATH."""
-    import distutils.spawn
-    from distutils.spawn import find_executable
-    return find_executable(name) is not None
+    return shutil.which(name) is not None
 
 # return the first command of the list that can be found on the OS
 def get_command_of(commands):


reticulatedpines:
Fixed by 3f24042a4dfbfba1c4cbfafede50e92a770ed3a9.

Simplified the code some, and fixed an unrelated error that hid error output in some cases of build failure.

PR #288, opened 2026-08-01
gcc 16: silence set-but-unused warnings on kept counters

coon42:

Here is the AIs report:

### What
  Annotate eleven intentionally-unread variables with the existing `UNUSED_ATTR`
  macro from `src/compiler.h`. No statements are changed.

  ### Why
  gcc 16 broadened `-Wunused-but-set-variable`: it no longer exempts
  `volatile`-qualified variables, and it flags plain locals that 15 let through.
  ML builds with `-Werror`, so each one stops the build:

  ../../src/reboot.c:155:25: error: variable 'k' set but not used [-Werror=unused-but-set-variable=]

  Note the trailing `=` — that is the option's argument-taking form, which gcc 15
  does not recognise at all. Confirmed by compiling the same code on 15.2.1 (no
  warning) and 16.1.0 (error).

  ### Why annotated rather than deleted
  Three of these are read by code that compiles out or is not built by default, so
  deleting them would break a debug build rather than tidy it:

  - `tasks.h` — `TASK_LOOP`'s `k` is read by `idle_led_blink_step(k)` in
    `zebra.c`. 43 call sites ignore it, and those are what warn.
  - `ml-cbr.c` — `count` is read by `dbg_printf()`, which is `do {} while(0)`
    unless CBR debugging is enabled.
  - `mlv_rec.c` — `writing_slots` is read only by a commented-out `trace_write`.

  The rest (`reboot.c` / `reboot-all.c` `busy_wait`'s `k`, `module.c`,
  `rbf_font.c`, `greenscreen.c`, `init.c`, `lvinfo.c`, `selftest.c`) are counters
  kept for documentation and possible future use.

  Worth noting for reviewers: `lvinfo.c`'s `used_items` looks like a missing
  division, because the two sibling functions do `extra_spacing / used_items`.
  `lvinfo_squeeze_space` deliberately doesn't — it works from `used_width` and
  discards by priority. A comment now records that, so it doesn't get "fixed"
  later.

  ### Risk
  Purely additive — the diff adds only an attribute and comments, so camera
  behaviour is unaffected on every generation.

  ### Testing
  Built `platform/200D.101` with `arm-none-eabi-gcc 15.2.1`: all 23 default
  modules and `magiclantern.zip`. Earlier revisions of the `reboot.c` and
  `tasks.h` changes were also built with `arm-none-eabi-gcc 16.1.0`, which is
  where the warnings were first reported. Not tested on a physical camera.

  To enumerate this warning class in one pass rather than one build at a time:

  ```sh
  cd platform/200D.101
  make clean && make FATAL_WARNINGS=n -k -j$(nproc) 2>&1 | grep -E "unused-but-set|warning:" | sort -u
  ```

  🤖 Generated with Claude Code (https://claude.com/claude-code)

### DIFF
--- a/modules/raw_video/mlv_rec/mlv_rec.c
+++ b/modules/raw_video/mlv_rec/mlv_rec.c
@@ -3436,7 +3436,8 @@ static void raw_video_rec_task()
         }
 
         uint32_t used_slots = 0;
-        uint32_t writing_slots = 0;
+        /* kept for the commented-out trace_write below */
+        uint32_t UNUSED_ATTR(writing_slots) = 0;
         uint32_t queued_writes = 0;
 
         /* this will enable the vsync CBR and the other task(s) */
--- a/modules/selftest/selftest.c
+++ b/modules/selftest/selftest.c
@@ -1375,7 +1375,8 @@ static void thread_test_task(void* arg)
 
 static void rpc_test_task(void* unused)
 {
-    uint32_t loops = 0;
+    /* counts the RPC pings sent; kept for documentation and future use */
+    uint32_t UNUSED_ATTR(loops) = 0;
 
     ml_rpc_verbose(1);
     while(1)
--- a/src/greenscreen.c
+++ b/src/greenscreen.c
@@ -77,7 +77,7 @@ void green_screen_step()
       uint16_t * const b_row = (uint16_t*)( bm        + BM_R(y)       );  // 2 pixels
       uint16_t * const m_row = (uint16_t*)( bm_mirror + BM_R(y)       );  // 2 pixels
       
-      uint8_t* lvp; // that's a moving pointer through lv vram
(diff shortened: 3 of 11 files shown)


reticulatedpines:
I don't like UNUSED_ATTR.  It's much better to fix the problem rather than suppress a valid warning.  I don't have gcc 16 to test, but I would also assume if you use a variable that is marked UNUSED, that's another, different error.  As some of these are conditionally unused based on build profile (e.g. some debug flags), if I'm right with that assumption the change will only cause problems in those configs.

This can wait until my system picks up gcc 16, then it will be faster for me to find appropriate fixes.

(possibly relevant: take a look at the recently introduced ARM_BINPATH env var, which allows selecting a specified gcc cross compiler)

coon42:
The AI simply removed them. The reason I have chosen to keep them and use `UNUSED_ATTR()` instead was, that some of the unused vars look like they are documentational. But fine, let's wait until you upgrade to gcc 16.

PR #291, opened 2026-08-04
feat: add SD overclocking support for DIGIC 4 SD bodies

wavesoft:
This PR adds a new **Debug** menu option that appears on DIGIC 4 bodies with an SD card and enables overclocking the SD controller itself.

Normally SDHC cards operate at a maximum of 48 MHz in High-Speed mode. Newer UHS-class cards are *capable* of higher clocks, but they usually do so in a UHS configuration at 1.8 V for better signal integrity. This change stays in the High-Speed path and instead raises the SD root clock (PLL1) above Canon's 48 MHz setting. PLL1 seems not to clock any other domains of the camera, making this a relatively safe addition.

The menu offers 48 / 72 / 96 / 144 MHz. An overclock at 72 MHz gives about 1.5× throughput, and at 96 MHz about 2×. Signal integrity is not guaranteed the higher you go - 144 MHz is labeled DANGER and wedged the SD driver on the card used for testing (with a dirty-flag recovery path that resets to stock after a failed boot). But you might have better luck with yours.

Settings are applied at boot after DryOS finishes card bring-up; a restart is required after changing the menu. Post-apply write/read verification refuses to keep a broken clock, and the feature self-disables if PLL1 is not the expected stock word at init (so it stays inert on unexpected silicon).

Benchmarked ~40 MB/s on an EOS 600D (with a SanDisk Extreme 128GB):

600D-SD

## Notes
- Compile-gated to `CONFIG_DIGIC_IV && !CONFIG_CF_SLOT` (500D / 550D / 600D / 60D / 1100D). Distinct from the existing Digic 5+ `sd_uhs` module.
- Only validated PLL divider words are used; off-grid words can hard-hang the camera.

PR #292, opened 2026-08-06
feat: enable LJ92 lossless compression on Digic 4 bodies

wavesoft:
> ℹ️ This PR is building on top of #291 

This enables Canon's Mem1 -> JPCORE lossless path on the 600D so `mlv_lite` can record **14-bit / 12-bit / 8..11-bit lossless**, plus an optional low-FPS soft preview while JPCORE freezes Evf.

Split into two commits so soft preview can be reverted independently if it causes issues.

### Test Builds

Here are some test builds:

* For EOS 600D / Rebel T3i : [magiclantern-eos600d.zip](https://github.com/user-attachments/files/30797215/magiclantern-eos600d.zip)

### DIGIC 4 lossless core

- Wire `ProcessTwoInTwoOutJpegPath` in `silent/lossless` (split Start, geometry patch, stall-detector completion, LV save/restore, OBWB gain for reduced depths).
- `mlv_lite`: encodable width (`4 mod 8`), overread rows, RAWI level mapping, 6-frame encode warm-up, clean compression-fault stop/drain.

- `raw.c` / `state-object.c`: DIGIC 4 digital-gain fix; Evf EngineError trim so borrowing JPCORE cannot ASSERT the camera.

VRAM0

### Soft preview (separate commit)

- Private low-prio task paints RAW→YUV at ~5 Hz gray / ~2.5 Hz color while recording lossless.
- Skips on real write backlog; OFF restores frozen-LV behaviour.

### DIFF
--- a/modules/raw_video/mlv_lite/mlv_lite.c
+++ b/modules/raw_video/mlv_lite/mlv_lite.c
@@ -364,6 +364,56 @@ static volatile                 uint32_t skip_frames = 0;
 /* for compress_task */
 static struct msg_queue * compress_mq = 0;
 
+/* In-flight compression jobs and first failed frame. Used to drain the queue
+ * cleanly on a compression fault and to drop any later frames already captured. */
+static volatile unsigned int compression_pending_jobs = 0;
+static volatile int compression_first_failed_frame = INT_MAX;
+
+#define COMPRESS_JOB_TRACKED 0x40000000
+#define COMPRESS_JOB_WARMUP  0x20000000
+
+/* Frames the encoder runs over and throws away before a compressed clip starts.
+ *
+ * The first encode of a clip swaps a set of LiveView registers over to our
+ * geometry, and LiveView needs a few frames to relock: in every DIGIC 4 lossless
+ * clip VIDF 1 to 4 came out shredded into bands of about sixteen rows while the
+ * vsync interval wobbled between 37.8 and 45.1 ms, and both the damage and the
+ * wobble stopped at VIDF 5. Uncompressed clips, which never touch those
+ * registers, are clean from VIDF 0.
+ *
+ * Skipping the frames outright would not help, because it is the first encode
+ * that disturbs LiveView - the transient would just move to wherever the first
+ * kept frame is. So encode over the unstable window with the results binned. */
+#define COMPRESS_WARMUP_FRAMES  6
+static volatile int warmup_frames = 0;
+
+static int count_free_slots(void);
+
+static void compression_job_admit(void)
+{
+    uint32_t old_int = cli();
+    compression_pending_jobs++;
+    sei(old_int);
+}
+
+static void compression_job_retire(void)
+{
+    uint32_t old_int = cli();
+    if (compression_pending_jobs)
+        compression_pending_jobs--;
+    sei(old_int);
+}
+
+static void compression_record_fault(int frame_number, int code)
+{
+    uint32_t old_int = cli();
+    if (compression_first_failed_frame == INT_MAX)
+        compression_first_failed_frame = frame_number;
+    buffer_full = 1;
+    sei(old_int);
+    (void) code;
+}
+
 static GUARDED_BY(RawRecTask)   mlv_file_hdr_t file_hdr[CARD_COUNT];
 static GUARDED_BY(RawRecTask)   mlv_rawi_hdr_t rawi_hdr;
 static GUARDED_BY(RawRecTask)   mlv_rawc_hdr_t rawc_hdr;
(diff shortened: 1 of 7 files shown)


wavesoft:
Something important I just realized: before starting recording you must select maximum video recording resolution from the default canon menu. If you had it on 720p or 480p it will fail to record.

mungewell:
Can you please explain the horizontal limit of 1724 pixels? If possible can we have this set to (slightly) above 1920, so that we can fill 1920x1080 frame/crop down in post?

Reasoning: We can record 1724x518-VC (2:1 ratio) 14bit-compressed continuously. 

_'VC' is the trick where Canon menu is set to 1280x720@60, and then we override FPS to 29.97 and shoot raw with the image 'Vertically-Compressed'. MLVApp automatically sees this and expands when processing (or you can expand in editing)._

EDIT: Updated dmo Video...
https://www.youtube.com/watch?v=dzxLlQBSe4w

wavesoft:
> Can you please explain the horizontal limit of 1724 pixels?

So, the sensor itself can do 5184×3456 in photo mode, however when operating in LiveView (the stream the RAW video taps into) it does 3x column binning and 3x row skipping, resulting to 1728x1152. 

The slightly smaller 1724 comes from the compression path. Digic 4’s SNDPAS (image-engine register block) feeder always advances 20 samples past the encoded width, so RD1 (JPCore input of EDMAC) must be feed width+20 per source row to keep rows aligned - and that feed is an even number of bytes only when width = 4 (mod 8).

> If possible can we have this set to (slightly) above 1920, so that we can fill 1920x1080 frame/crop down in post?

Unfortunately (with the knowledge I collected so far) Canon's internal pipeline (sensor clocks, ADTG, EDMAC, DRAM) is sized for the small stream (~1888×1181), so that won't be possible.

> We can record 1724x518-VC (2:1 ratio) 14bit-compressed continuously.

Oh I wasn't aware of this trick. I need to reproduce it first and then try to see if I can make it work 😄

mungewell:
On the above demo video, the clip with 'the dog on the left' there is a fair amount of 'shimmer' on the forest floor. I don't think that this is specifically an issue with the raw compression. It looks like the '-VC' hack show it worse... I took several format clips in each location, which also show a comparison between different raw formats and in camera H264/MP4.

It's a bit masked with the You-Tube compression. Looks worse in the files on my PC.
https://youtu.be/rxbWxXeaZZ0

Any suggestions on how this can be masked? Option in MLV-App?

mungewell:
I did some testing on a tripod, and noticed that the crop shown on monitor changes during the '14bit lossless' preview.
Screenshot_2026-08-08_20-54-32

Neither the 'before' or 'during' crop actually matches the output file....

wavesoft:
> Any suggestions on how this can be masked? Option in MLV-App?

This feels like aliasing. Haven't played much with MLV app options, but check this perhaps: https://youtu.be/dMfYUJRQ27Q?si=_ZlDnYOPkZL8JX3a&t=270 it's for EOS M, but I think the slight blurring you will get from Chroma Blur Radius _might_ help.

> I did some testing on a tripod, and noticed that the crop shown on monitor changes during the '14bit lossless' preview.

That's expected, because while the compression engine is busy, the live preview is turned off. So I added a low-quality, low-CPU overhead, software-only preview option (some choices available on "Soft Preview" inside the raw video menu). That's probably what you are looking at.

reticulatedpines:
> So, the sensor itself can do 5184×3456 in photo mode, however when operating in LiveView (the stream the RAW video taps into) it does 3x column binning and 3x row skipping, resulting to 1728x1152.

On most (all?) cams, when you select 5x or 10x zoom, the captured area for LV changes, and is 1:1, not 3:3 (presumably so the user can pan around anywhere with joystick).  You can then crop out a target region to capture as raw video.

crop_rec module has code and GUI for this.  The GUI is quite confusing.

mungewell:
I figured out what is happening in my crop example/question.

ML is adding the crop to the Overlay before recording, and once the recording starts 'soft-preview' takes over. **BUT** 'soft-preview' draws on the whole monitor whilst Overlay is still active and still showing a crop - so 'soft-preview' is then **further cropped**.... 

I have half-shutter assigned to clear overlay, if I do this during recording I see the whole screen and 'soft-preview' as it is intended to be. This view matches recorded video.

EDIT: On using x5 (x10) zoom, these temporarily clear the crop. Perhaps 'soft-preview' can do this too...

wavesoft:
> ML is adding the crop to the Overlay before recording, and once the recording starts 'soft-preview' takes over. BUT 'soft-preview' draws on the whole monitor whilst Overlay is still active and still showing a crop - so 'soft-preview' is then further cropped....

Oh that's a really good find. I don't think I tested the x5/x10 zoom much 🤔 Let me iterate over this...

reticulatedpines:
> I don't think I tested the x5/x10 zoom much 🤔 Let me iterate over this...

If you log the size of the EDMAC transfer while you change modes, it should be clear.  display_buf.py is also useful.

PR #293, opened 2026-08-10
developer_guide: write the MPU section

ssubbotin:
The MPU section of the developer guide was a placeholder reading "ML relevant
information re MPU goes here." This fills it in from what is already recorded in
the tree.

Covered:

- the SIO3 link and MREQ handshake, and where the per-cam MMIO addresses are
  listed (`hw/eos/model_list.c` in qemu-eos)
- the Canon side function names, and which cams they are currently located on,
  with the note that only the 100D is D45
- how to find them on a new cam: the `"dwSize < TXBD_DATA_SIZE"` assertion
  string, `InitializeIntercom` leading to `mpu_recv_cbr`, and the three
  `allocRingBuffer` sizes, pointing at kitor's notes in
  `platform/R.180/stubs.S` for the detail
- the two gotchas: 5D4 has two `mpu_send` functions, and on D45 the addresses
  are RAM addresses because the code is relocated at boot
- the message format, and the class/id mapping in `known_spells.py`
- the seven `0xCCCCCCCC` group messages, which are the largest in the tables and
  whose structure is not worked out
- the two MPU parts, and that `mpu_dump` only works on the older TMP19A43

I also tried to be explicit about what is *not* known, because the ~300 entry
property mapping can suggest more understanding than we have. We can usually say
which property a message carries; we generally cannot say what its bytes mean.
That is the reason qemu-eos replays recorded transcripts instead of generating
messages, and it seemed worth stating in the guide rather than leaving a reader
to infer it.

Everything here comes from comments in the tree, the stub files, and
`hw/eos/mpu.c`, with the source named in each case so it can be checked or
corrected.

### Style

I matched the existing sections rather than introducing anything new: `###`
headings only (no other section in the guide uses `####`), flowing prose rather
than subsections, escaped underscores, and roughly 95 character lines. The
section is 72 lines, against 41 for EDMAC and 60 for ADTG.

Regenerated with `generate_guide.py` to confirm it renders; the HTML output is
correct and the escaped underscores and angle brackets come out as intended.
Happy to trim or restructure if it reads as too long for its slot.


reticulatedpines:
Hi - I don't recognise your name from the forum or Discord.  What's your background dealing with MPU hardware?

Some of the information in the commit is useful, but it's nearly all information about software, and section 2 is supposed to be for hardware.  Obviously, the developer guide is in early stages and not well structured yet, but my intention was for detailed software information to occur after an introduction to the hardware.

Some of the paras in the commit are kind of unrelated.  They vaguely relate to MPU, but don't have a good narrative flow for a developer guide.  Are these LLM written possibly?  They're not as good at keeping track of long range info.

Probably for these technical sections, we need more discussion before considering a PR.  We'd want to validate any new info, make sure any assertions are correct on listed cams, etc.  If you're on the forum or Discord, those are better places than github for detailed discussion.

reticulatedpines:
Closing due to no response.

Issue #128, opened 2024-01-24
Rename "contrib" to "tools", tidy up


reticulatedpines:
contrib was never a great name, it doesn't tell you anything about what it should do.  Most things in there vaguely fit under "tools".  Rename to that, and split out the contributer named subdirs as appropriate.

Also, delete qemu subdir (after checking for anything that isn't already preserved elsewhere given the qemu-eos repo).

Add 14_to_16.c and any other useful things.

reticulatedpines:
Done.

PR #294, opened 2026-08-15
6D2: override the 29:59 MOV/MP4 recording limit

ServerDestroyer:
Adds the two `MVR_TIME_LIMIT_*` constants and enables `FEATURE_OVERRIDE_MOVIE_30_MIN_LIMIT` for the 6D2 (1.1.1), following the 200D implementation.

**How the addresses were found:** searched ROM0 for the exact `.old_value` constants `movtweaks.c` asserts against — 1799000 (29m59s) and 449000 (7m29s). Each byte pattern occurs exactly once in the whole 32 MiB image, adjacent and in that order, so there is no competing candidate. Both are reached from a two-armed getter at `0xe042fee8` (the `beq` at `0xe042fef0` selects normal vs high FPS; each arm is an `ldr r0, [pc, #128]` whose computed T1 literal target resolves to these addresses). Verified by decoding the instructions, not by byte pattern-matching alone.

**Tested on a real 6D2 (fw 1.1.1):** limit set to 1 min → recording stops on its own at ~60 s. `apply_patches()` accepted both patches without `E_PATCH_OLD_VALUE_MISMATCH`, independently confirming the `.old_value` constants against this body's ROM. Also double-checked by two consecutive ~448 MB MP4s landing within 0.2% of each other at the 180-min setting test.

### DIFF
--- a/platform/6D2.111/consts.h
+++ b/platform/6D2.111/consts.h
@@ -131,6 +131,16 @@ extern int winsys_bmp_dirty_bit_neg;
 //#define MVR_LAST_FRAME_SIZE (*(int*)(512 + MVR_752_STRUCT))
 #define MVR_BYTES_WRITTEN MEM((212 + MVR_190_STRUCT))
 
+// Recording time limits, in ms, in a literal pool in ROM0.
+// Both are read by a two-armed getter at 0xe042fee8: the beq at 0xe042fef0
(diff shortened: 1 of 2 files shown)


reticulatedpines:
Closed by 41025bec9ef562a2198790567f12f2e12498ec9b (I cleaned up the comments a lot, otherwise unchanged)

PR #296, opened 2026-08-15
log-d678: don't spin forever when the log buffer allocation fails

ServerDestroyer:
`log_start()` spins in `while (!buf);` if the initial log-buffer allocation fails. Since this runs from `boot_post_init_task`, an allocation failure presents as an apparently-bricked camera — black screen, no response, nothing to debug with.

The failure is reachable in practice: we hit it on the 6D2 while investigating startup logging (allocation failing at `log_start()` time under certain feature combinations). With this change the logger degrades gracefully — sets `buf_size = 0` and returns, so the camera boots normally, just without a startup log.

Also fixes a gcc 15 build error in the same file (`task_name_padded[11]` has no room for the NUL terminator; `-Werror=unterminated-string-initialization`).

### DIFF
--- a/src/log-d678.c
+++ b/src/log-d678.c
@@ -313,7 +313,17 @@ void log_start()
 #endif
     qprintf("Logging buffer: %X - %X\n", buf, buf + buf_size - 1);
     qprintf("Free memory: %X\n", GetFreeMemForAllocateMemory());
-    while (!buf);
+    if (!buf)
+    {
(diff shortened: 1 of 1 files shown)


reticulatedpines:
Why is it preferable to proceed when logging is impossible, with a build explictly opting into logging?  (note that the qprintf calls before the while(!buf) will show the cause of the problem in qemu build using CONFIG_QEMU=y).

reticulatedpines:
Closing due to no response

PR #313, opened 2026-09-17
EOSM: add opt-in read-only USB LiveView preview

gdemontalivet:
# EOSM: add opt-in read-only USB LiveView preview

This adds a bounded PTP operation (`0x9997`) that copies the current original-EOS-M LiveView UYVY422 buffer to a USB host. It is deliberately read-only, disabled by default, and guarded to EOS M builds; it does not enable CHDK's arbitrary-memory PTP API.

## Purpose

I use an EOS M to camera-scan a large collection of slides and negatives. LiveView lets the host detect when the film/camera has stopped moving, inspect framing and exposure, and then coordinate a full-resolution RAW capture before advancing to the next frame. The preview is for positioning and automation; it is not used instead of the final CR2 photograph.

Canon's normal remote-viewfinder request is not available on this original EOS M: the camera returns PTP response `0x2005` (operation not supported). HDMI capture would provide a preview but adds another cable and capture device, and would not share the existing USB control path.

## Current PR scope

The branch contains the first read-only implementation: protocol version 1, a bounded 720×480 UYVY frame command, a single-frame libusb receiver, and an SDL continuous viewer. It does not start LiveView, change camera settings, trigger the shutter, or expose arbitrary camera memory. Remote-control experiments remain outside this PR.

On one original EOS M running Canon firmware 2.0.2, the version command, 720×480 single-frame transfer, and continuous viewer worked. A later same-session measurement produced about 14.3–16.1 fps.

## Follow-up HD-buffer experiment

After the question about PTP picture transfer and higher resolution, I tested a follow-up custom PTP command using Magic Lantern's existing `get_yuv422_hd_vram()` buffer. On the same physical camera it returned a valid continuously updating 960×639 UYVY LiveView image: 613,440 pixels versus 345,600 at 720×480, or 1.78× the pixel count. Sustained HD preview measured about 7.5–9.5 fps because each uncompressed frame is 1,226,880 bytes rather than 691,200 bytes.

The experiment exposes both buffers as separate bounded subcommands, so a host can choose between lower-resolution/higher-rate 720×480 and higher-resolution/lower-rate 960×639. The earlier implementation was already transported over PTP; the improvement comes from exposing the higher-resolution ML buffer, not from replacing a non-PTP transport. This follow-up v6 implementation is not currently part of the PR branch.

## Validation

- Full `platform/EOSM.202` build with `CONFIG_PTP=y CONFIG_PTP_PREVIEW=y`.
- Both host examples compile with Clang and warnings treated as errors.
- Synthetic signed-chroma UYVY decoder test passes.
- Physical-camera tests completed for the 720×480 PR implementation and the separate 960×639 experiment.
- Existing upstream build warnings appeared, but no new camera-handler warnings were observed.

## LLM disclosure

I used OpenAI Codex extensively to generate and guide the current PR code, documentation, host tools, debugging, and the follow-up HD experiment. I am not presenting the code as independently handwritten. In light of this repository's LLM policy, please treat the PR as an experimental demonstration and technical discussion rather than a merge-ready contribution. If the approach is useful, I can use what I learned to produce an independently understood and rewritten implementation that complies with the project policy.


reticulatedpines:
Hi - have you read our LLM policy?  https://github.com/reticulatedpines/magiclantern_simplified/blob/e15e0c2d2bece6285dde0e4528c364980c348018/LLM_POLICY.md

We cannot accept this code at present, it is clearly LLM generated.

Also, no explanation is given for why you want this code, what it is for, etc.  I can't review a PR without knowing that.  I also can't answer your question, for a similar reason; I don't know what your experiment is for.

gdemontalivet:
Hi, sorry i thought this could be beneficial to others, didn’t realised there was an llm policy. I have an EOS M1 and this adds the capability to support live view over usb . The aim is to speed up negative scanning by automating it with remote control over usb as well. 
apoloogies again

reticulatedpines:
Yup, it might be useful for others!  But I need to know what you want it for, before I can work that out.  We also need to make sure the entire codebase makes sense for all cameras we support - that's about 30 different models.

I don't own an M, my understanding is the official USB support is not good.  That said, it should support PTP picture transfer.  Here it sounds like you're automating taking pictures with a scanner involved?  PTP could do that with the same wires connected, and you'd get higher resolution images.  HDMI capture is also an option, though one that requires (cheap) additional hardware.

Do you have specific needs where these pre-existing solutions wouldn't work?

WalterSchulz:
Actually, IMO: No, it doesn't make sense for all other cameras. Because M got a weird USB feature set by Canon not supporting liveview over USB. All other (at least Digic 4 and up) do support this. 

We had this before: Canon EOS Webcam Utility, EOS Utillity, SparkoCam and gphoto2 won't do anything about liveview via USB because cam doesn't allow it.

This feature will be there for EOS M and EOS M only.

reticulatedpines:
Sorry, I was ambiguous: perhaps useful for other people (with M's), not other models.

gdemontalivet:
to explain what I am doing; I have a lot of slides and negative to scan with my eos m. so now with live view through magic lantern and full remote support as well (I have it in another pull request ready ;-) , I now have a script which detects when the camera stops moving, takes a picture of the negative, then I move to the next negative, when it stops moving, it takes a picture again, and so on. I can also automate to ge the histogram, change the adjustments, take the photo and go to the next one!

reticulatedpines:
Sounds fun to watch!  Can you answer my questions from earlier?  It feels like you could do the same thing with existing functionality.

gdemontalivet:
You were right that PTP can carry a higher-resolution LiveView over the same USB connection. We tested it on the original EOS M by adding a bounded Magic Lantern PTP command that reads the camera's HD LiveView buffer. It delivers 960×639 versus 720×480—about 1.78× more pixels.

The tradeoff is speed: the raw UYVY stream runs around 7.5–9.5 fps, compared with 14.3–16.1 fps at 720×480, because each frame contains 1.78× more data. The extension exposes both buffers as separate commands, so the viewer can select either lower resolution/higher frame rate or higher resolution/lower frame rate.

One qualification: Canon's standard remote-viewfinder PTP opcode still returns “operation not supported” on the original EOS M. Our working solution is a custom ML PTP extension, and the previous 720×480 implementation was also transported over PTP. So PTP was not the missing transport—the improvement came from exposing ML's higher-resolution LiveView buffer through it. The LiveView is only used for positioning and automation; the scanner still triggers and saves a full-resolution CR2 for each negative.

For transparency under the project's policy: I used OpenAI Codex extensively to generate and guide this implementation and test work. I have updated the PR description with that disclosure and with the measured results. I have not pushed the larger follow-up implementation into the PR branch as if it were independently handwritten; the v6 result is reported as an experiment for discussion.

reticulatedpines:
I wasn't suggesting using LV via PTP, I was saying one route would be to use *stock* PTP functionality to transfer a full image from the cam.  You can also trigger image capture with stock PTP.  gphoto2 should be capable of both these things, although I don't have an M to test against.

To restate my core question: given you can take and retrieve a picture over USB *without ML at all*, can you do the main thing you want (automating scanning) with existing functionality?  Note this is a separate question from whether you can do what you want *well*.

I'm trying to work out your actual requirements, so we can determine what the best approach is to solve whatever problem you have.  There are at least three existing routes: stock PTP, Lua scripting on cam (no USB cable required!), CHDK_PTP optional build config.  Do we really need a fourth?  Why is it worth the maintenance cost?

gdemontalivet:
Thanks—that distinction helps. I understand now that you meant using stock PTP to trigger and retrieve the final full-resolution photograph, not using PTP for LiveView.

I tested that route on the physical original EOS M / Canon firmware 2.0.2 with gphoto2 2.5.32 and libgphoto2 2.5.34. Stock PTP file access works: gphoto2 lists the CR2 files on the card and can download them. Capture did not work on this camera:

- `gphoto2 --abilities` reports `Capture not supported by the driver`.
- The live camera summary reports `No Image Capture, No Open Capture, No vendor specific capture`.
- One `gphoto2 --capture-image` attempt returned `camera does not support generic capture` / `Unsupported operation`.
- A subsequent file listing confirmed that no new image was created.

ML was running during that test, although our extension only adds a separate vendor opcode and does not replace Canon's standard PTP capability response. I can repeat it with ML completely disabled as a strict control.

There is also a separate LiveView requirement. The original EOS M does not expose LiveView over stock USB/PTP, and the existing ML build did not add it. The normal Canon viewfinder request returns `0x2005` (`OperationNotSupported`). Our change is what made an updating USB preview available. Therefore, even if stock PTP capture and retrieval worked, it would cover only the final-photo portion of the workflow—not the host-side framing, movement detection and exposure monitoring that use LiveView.

The scanning requirements are:

1. Let the Mac monitor framing and determine when the externally moved film carrier has stopped.
2. Inspect exposure/histogram information.
3. Trigger a full-resolution RAW capture.
4. Confirm that the capture was saved.
5. Retrieve the CR2, immediately or in a later batch.

LiveView is only used for monitoring and automation; it does not replace the final CR2.

There is a second use case we are exploring as well: streaming video and eventually audio over USB so the original EOS M can be used as a webcam. Experimental USB video now works at 720×480 and 960×639. USB audio streaming has not been implemented or validated yet, so that is a future goal rather than a capability claimed by this PR.

On-camera Lua could potentially handle some motion/stability detection, but it would need to replace or communicate with the host-side coordination of the film carrier, lighting and processing. CHDK_PTP is therefore the most relevant existing alternative, and I agree it should be evaluated before claiming that another PTP interface is justified.

So my current answer is: stock PTP handles retrieval but, on the EOS M tested, gphoto2 does not provide shutter triggering; stock USB also does not provide LiveView. I will test the no-ML baseline and CHDK_PTP next. If CHDK_PTP already covers the requirements reliably, the maintenance argument for a separate bounded interface becomes much weaker.

reticulatedpines:
Thanks for testing gphoto2.  Lack of support there doesn't prove the cam doesn't support it, sometimes they just haven't implemented things yet.  It's suggestive but not conclusive.  The M is well known to have weird USB support, so I'd be unsurprised if it doesn't allow this stock.

With CONFIG_CHDK, you would gain access to PTP_CHDK_CallFunction, which can be used to take a picture.  You can then download this pic and use it for motion detection - therefore I don't see a LiveView requirement.  As a bonus, if the pic is judged motionless, you can directly obtain the scan, since it is full res.

Since your approach is already a build time config, not intended for normal builds, extending PTP_CHDK would seem more appropriate (and likely no extension is needed, I think it should already do what you need).

We already have motion detection, exposure monitoring, and image capture in core ML code, all of which are likely accessible via Lua.  To me it sounds like you can drop the host side stuff and move almost everything onto camera, without needing to change ML code at all.

See especially:
https://builds.magiclantern.fm/lua_api/modules/camera.html

I don't see an obvious way to query exposure information, but I've never used the Lua interface, it may already be there, hiding.  If it isn't, we clearly know it in core code, see histogram.c.  Exposing this by extending the Lua interface with some read-only value would be easy.

Likewise motion detection, see FEATURE_MOTION_DETECT in shoot.c.  It will be possible to have a low-latency hook or callback to monitor state changes from Lua.

WalterSchulz:
Again: It is not about "gphoto2" and "yet". 
No (none) of the webcam apps/programs is able to connect to M via USB.
It doesn't work with EOS Webcam Utility, it doesn't work with SparkoCam (payware) or anything else requiring USB access. EOS Utility's liveview option: Not available for M. 
For example gimbals working with USB connected to cam have no support for M.

Tornado EOS doesn't support M either.

gdemontalivet:
I forgot to mention: the interest of LiveView also is to use an eos-m as a webcam, which I can now do with my eos-m1 using the camera as a microphone as well!

reticulatedpines:
That could also be done with PTP_CHDK, using basically the same approach you've tried; repeatedly polling for LV data.  The host client would differ a bit, and you wouldn't need any changes to ML code.

